PR #6, opened 2019-05-07
Fs on apr19 valeraonisko

valeraonisko:


src/style.css
```diff
@@ -0,0 +1,67 @@
...
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+
+}
+
+@keyframes revolve-min {
```
YuriiHoliuk: duplicated, you can use revolve

valeraonisko: Done). I was so happy that it moves, that even didn't think about duplications )

PR #26, opened 2019-05-11
Solution

SherJlok15:


src/style.css
```diff
@@ -0,0 +1,55 @@
...
+  width: 10px;
+  height: 300px;
+  position: absolute;
+  right: 10px;
+  top: -285px;
+  transform-origin: bottom center;
+  animation: rotate_hand 3600s linear infinite;
+}
+
+.minute_hand {
```
gestap: you confuse minutes and seconds

PR #28, opened 2019-05-13
Fs on apr19 andrii

Laliquenoare:


src/style.css
```diff
@@ -0,0 +1,69 @@
...
+
+.clock {
+  position: relative;
+  background: #4b9aaa;
+  border: 10px solid #4b9aaa;
+  border-radius: 50%;
+}
+
+.seconds {
+  content: "";
```
gestap: `content` works only with  pseudo-elements

PR #29, opened 2019-05-13
Vitalii Pshychenko

VitaliiPshychenko:


src/index.html
```diff
@@ -13,7 +13,10 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-
+  <div class="center">
```
YuriiHoliuk: better to create an additional container, but ok

PR #87, opened 2019-07-30
Fs jul19 dmytro m

Deman250:
https://deman250.github.io/layout_stop-watch/

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,67 @@
+.clock_container {
+  position: relative;
+  top: 100px;
+  left: 650px;
+  width: 700px;
+  height: 700px;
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,67 @@
+.clock_container {
+  position: relative;
+  top: 100px;
```
irenhh: why did you write top and left here?

Deman250: i want that my clock was in midle of screen

Deman250: clock moved

irenhh:
next time if you want to center the element you can use `margin: 0 auto` for containers and corresponding styles for children elements

PR #88, opened 2019-07-30
Solution_clock

ivansafiulin:


src/index.html
```diff
@@ -13,7 +13,12 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-
+  <div class="clock">
+      <div class="minute-hand"></div>
```
irenhh: fix the indentations

src/style.css
```diff
@@ -0,0 +1,56 @@
...
+  width: 29px;
+  height: 29px;
+  bottom: -16px;
+  right: -12px;
+  border-radius: 50%;
+  background-color: #4b9aaa;
+}
+
+@keyframes stopwatch {
+
```
irenhh: blank lines aren't necessary here

ivansafiulin: yes, senk you....understand it on media query from past homework

ivansafiulin: I will be more careful in the following homework

PR #96, opened 2019-07-31
sol

Taisiiapn:
[UPDATE DEMO LINK](https://taisiiapn.github.io/layout_stop-watch/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,56 @@
+.watch {
+  border-radius: 50%;
+  background: #fff;
+  height: 600px;
+  width: 600px;
+  position: relative;
+  margin: auto;
+}
+
+.watch__round {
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,56 @@
...
+  border-radius: 3px;
+}
+
+@keyframes rotate {
+  100% {
+    transform: rotateZ(360deg);
+  }
+}
+
+.minute {
```
danheim: Why is it taken out separately?

src/style.css
```diff
@@ -0,0 +1,56 @@
...
+  animation: rotate 3600s infinite linear;
+}
+
+.second {
```
danheim: The same

PR #102, opened 2019-08-01
Created clock with animation

Vyacheslav-P:
- [DEMO LINK](https://vyacheslav-p.github.io/layout_stop-watch/)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,0 +1,78 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.clock-position {
+  position: relative;
+  top: 100px;
+  left: 100px;
+}
+
+.clock {
+  width: 300px;
+  height: 300px;
+  border-radius: 50%;
+  background-color: rgb(233, 233, 250);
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,78 @@
+* {
+  margin: 0;
+  padding: 0;
+}
```
danheim: You sure you need these styles?

src/style.css
```diff
@@ -0,0 +1,78 @@
...
+@keyframes minute--animation {
+  from {
+    transform: rotate(-180deg);
+  }
+  to {
+    transform: rotate(180deg);
+  }
+}
+
+@keyframes second--animation {
```
danheim: You don't need two keyframes

src/style.css
Vyacheslav-P: No, not sure) fixed

src/style.css
Vyacheslav-P: fixed

PR #103, opened 2019-08-01
Fs jul19 volod - my lovely stopWatch

Volodsher:
Thanks a lot for your amazing job, TA!

https://volodsher.github.io/layout_stop-watch/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -13,7 +13,14 @@
   <link rel="stylesheet" href="./style.css">
 </head>
 <body>
-
+  <div class="watch-container">
+    <div class="watch minute-hand">
+    </div>
+    <div class="watch second-hand">
+    </div>
+    <div class="watch dot">
+    </div>
+  </div>
 <script type="text/javascript" src="./main.js"></script>
 </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/style.css
```diff
@@ -0,0 +1,46 @@
+.watch-container {
+  position: absolute;
+  left: 50%;
```
irenhh: no need to mention both left and right, something one is enough

Volodsher: Thanks for review, I corrected mistakes!

src/index.html
```diff
@@ -13,7 +13,14 @@
...
 </head>
 <body>
-
+  <div class="watch-container">
+    <div class="watch minute-hand">
+    </div>
+    <div class="watch second-hand">
+    </div>
+    <div class="watch dot">
+    </div>
```
danheim: No need to wrap the closing tag.
`<div class="watch dot"></div>`

src/style.css
```diff
@@ -0,0 +1,47 @@
...
+  height: 224px;
+  margin-top: 146px;
+  animation: rotation 3600s infinite;
+  animation-timing-function: linear;
+}
+
+@keyframes rotation {
+  from {
+    transform: rotate(0deg);
+    transform-origin: bottom;
```
danheim: You can place this property to `..hand` selectors.

PR #147, opened 2019-10-09
Simple development

mgrinko:


package.json
```diff
@@ -4,9 +4,12 @@
   "description": "Boilerplate for layout tasks with visual regression testing",
   "main": "src/main.js",
   "scripts": {
-    "deploy": "bash ./config/hooks/deploy.sh",
+    "deploy": "git add . && git commit -m 'solution' && git push origin develop && npm run publish",
```
mgrinko: `git add .` and `commit` are required to save changes in README file (it is outside the `/src` folder)

roman-mate: There is risk they'll be committing some crap into the repo, but I think that's fine.

PR #247, opened 2020-03-05
is ok

Vlad-Viddi:
- [DEMO LINK](https://Vlad-Viddi.github.io/layout_stop-watch/)


vpolets:
image
а центровочка то хромає)

Vlad-Viddi:
> image
> а центровочка то хромає)

I've been centered it, see the code, and I'm definitely sure it's should be correct. donno why it's looks like this :)

PR #274, opened 2020-04-29
solution

aleksey-10:
- [DEMO LINK](https://aleksey-10.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,64 @@
+$sw-size: 500px;
+
+.stopwatch {
+  width: $sw-size;
+  height: $sw-size;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,64 @@
...
+  width: 5px;
+  position: absolute;
+  z-index: 0;
+  border-radius: 2px;
+  animation: move-arrow 60s linear 1s infinite normal backwards;
+
+  &-sec {
+    height: $sw-size;
+
+    &::after {
```
YuriiHoliuk, line 33: `::before` and `::after` are pseudoelements and should be written with `::`

PR #305, opened 2020-07-05
Solution

oggyrock:
[DEMO LINK](https://oggyrock.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,68 @@
+$second_length: 200px;
+$second_thickness: 6px;
+$minute_length: 150px;
+$minute_thickness: 10px;
```
vkryvytskyy, line 4: would be nice move variables to `variables.scss` and import there

PR #309, opened 2020-07-08
solution

ArtemiiKab:
- [DEMO LINK](https://ArtemiiKab.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,66 @@
+$second_thickness: 10px;
+$second_length: 340px;
+$minute_thickness: 10px;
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,66 @@
...
+  from {
+    transform: rotate(0deg);
+    transform-origin: bottom center;
+  }
+
+  to {
+    transform: rotate(360deg);
+    transform-origin: bottom center;
+  }
+}
```
vsm2k17, line 66: you can use one animation in both cases, just change time

PR #320, opened 2020-07-11
solution

furude-rika:
- [DEMO LINK](https://furude-rika.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/watch.scss
+++ b/src/styles/watch.scss
@@ -1,0 +1,43 @@
+.watch {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: $minuteHandHeight * 2 + $emtywatchSpace;
+  height: $minuteHandHeight * 2 + $emtywatchSpace;
+  border: 10px solid black;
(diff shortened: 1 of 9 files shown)


src/styles/watch.scss
```diff
@@ -0,0 +1,43 @@
...
+  justify-content: center;
+  align-items: center;
+  width: $minuteHandHeight * 2 + $emtywatchSpace;
+  height: $minuteHandHeight * 2 + $emtywatchSpace;
+  border: 10px solid black;
+  border-radius: ($minuteHandHeight * 2 + $emtywatchSpace) / 2;
+  background: linear-gradient(to right, #ff4e50, #f9d423);
+
+  &__minute-hand {
+    position: relative;
```
nanccyy, line 12: for these two hands you can add properties such as position, width, border-radius, transform-origin into a mixin like "hands-basic-configuration".

irenhh: good point, use it

PR #384, opened 2020-11-12
Added clock with animation

AlexysUa:
- [DEMO LINK](https://alexysua.github.io/layout_stop-watch/)
- There are no tests for this task so use npm run lint command instead of npm test


vsm2k17 (approved):
good job

PR #489, opened 2021-04-02
Solution

igor-stasiv:
- [DEMO LINK](https://igor-stasiv.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -18,10 +18,10 @@
   </head>
 
   <body>
-    <main class="watch">
-      <div class="minute-clockwise"></div>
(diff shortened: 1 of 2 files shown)


Burunduk003 (approved):
Great!

VinnieJ-2k20 (changes requested):
Let's follow BEM methodology.
If an item has absolute positioning, it should be an element of some block.
That refers to your clock center and minute and second hands.

Also, read about animation shorthand property

VinnieJ-2k20 (approved):
Good job 👍

PR #496, opened 2021-04-04
Solution

iryna-vys:
[DEMO LINK](https://iryna-vys.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/stopwatch.scss
+++ b/src/styles/blocks/stopwatch.scss
@@ -1,0 +1,45 @@
+.stopwatch {
+  display: flex;
+  flex-grow: 1;
+  justify-content: center;
+  align-items: center;
+  position: relative;
+
+  &__center {
+    position: absolute;
+    z-index: 1;
+    height: 30px;
+    width: 30px;
+    background-color: #4b9aaa;
+    border-radius: 50%;
+  }
+
+  &__hand {
+    position: absolute;
+    width: 12px;
+    border-radius: 6px;
+
+    &--second {
+      height: 30%;
+      background-color: #d15559;
+      bottom: 50%;
+      transform-origin: bottom;
+      animation-name: move;
+      animation-duration: 60s;
+      animation-timing-function: steps(60);
+      animation-iteration-count: infinite;
+    }
+
+    &--minute {
+      height: 50%;
+      background-color: #4b9aaa;
+      top: 0;
+      z-index: -1;
+      transform-origin: bottom;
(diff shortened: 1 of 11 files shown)


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,45 @@
...
+
+    &--second {
+      height: 30%;
+      background-color: #d15559;
+      bottom: 50%;
+      transform-origin: bottom;
+      animation-name: move;
+      animation-duration: 60s;
+      animation-timing-function: steps(60);
+      animation-iteration-count: infinite;
```
IvanGrekov, line 30: If you use such form of writing animation you can use mixin or placeholder to reduce code's strings

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    &--minute {
+      height: 50%;
+      background-color: #4b9aaa;
```
IvanGrekov, line 35: You can create variable for thic color

IvanGrekov (approved):
Good job! But I left a couple of comments. 

Also I think second hand of watch this is hand that longer and shorter hand is minute hand. In your solution it is vice versa)

VinnieJ-2k20 (changes requested):
Ivan is correct, second hand is usually longer than minute one.
Fix it.

Also, read about animation shorthand property:

iryna-vys:
Misha said, that he liked when properties were exactly visible. So I use it and I like it too.

VinnieJ-2k20:
> Misha said, that he liked when properties were exactly visible. So I use it and I like it too.

No problem, it's not a mistake.

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,45 @@
...
+      height: 30%;
+      background-color: #d15559;
+      bottom: 50%;
+      z-index: -1;
+      transform-origin: bottom;
+      animation-name: move;
+      animation-duration: 3600s;
+      animation-timing-function: linear;
+      animation-iteration-count: infinite;
+    }
```
VinnieJ-2k20, line 32: Don't repeat yourself.
It's better to put all styles that are same for `--minute` and `--second` modifiers under `__hand` selector.

VinnieJ-2k20 (approved):
Good job 👍 

Left some comments with advice about code optimization.

PR #503, opened 2021-04-05
solution by ched

ched88xtc:
- [DEMO LINK](https://ched88xtc.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,63 @@
+* {
+  margin: 0;
+  padding: 0;
+}
+
+.watch-container {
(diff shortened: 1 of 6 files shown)


ViPiven (approved):

src/styles/main.scss
```diff
@@ -0,0 +1,63 @@
+* {
+  margin: 0;
```
aionova, line 2: don't remove margins and paddings for all the selectors. Use reset.scss for that

aionova (approved):
:fire::fire::fire:

PR #512, opened 2021-04-06
solution

KapToH337:
[DEMO LINK](https://KapToH337.github.io/layout_stop-watch/)


aionova (approved):
Cool animations!
But you were supposed to make a minute and second hands of the clock. It takes a second hand 1 minute to make a full circle, and it takes an hour to make a full circle for the minute hand.

PR #549, opened 2021-05-31
stopwatch homework

yatania:
- [DEMO LINK](https://yatania.github.io/layout_stop-watch/)


aionova (approved):
Good job!
But you could try to use placeholders or mixins indeed

PR #592, opened 2021-06-22
Task done

Peephie:
https://peephie.github.io/layout_stop-watch/


mykola-yarchuk (dismissed):
https://user-images.githubusercontent.com/35385361/122937525-b91ba680-d37a-11eb-83d0-b717f87f2921.mov

should work as a stopwatch, should be a second hand and a minute hand

here an example

https://user-images.githubusercontent.com/35385361/122938329-5676da80-d37b-11eb-95b6-fa23fe81af47.mov

vpolets (approved):

PR #605, opened 2021-07-12
Clock done

ivanchenko-kirill:
- [DEMO LINK](https://ivanchenko-kirill.github.io/layout_stop-watch/)
I added alternative linear animation in comments, because there is no information in the task about which one we should use.


lera-yaskevych (commented):
The code is wonderful! I have nothing to suggest 👍

Denchepornyuk (approved):

PR #607, opened 2021-07-13
Clock done

illiarerun:
- [DEMO LINK](https://illiarerun.github.io/layout_stop-watch/)


lera-yaskevych (commented):
The code is really good!
I guess I can only suggest you place the watch in the center of the page, just for the sake of practicing BEM and flexbox positioning, since there is no such requirement in the list, so it's up to you)

illiarerun:
Thanks for your review! I should've thought about centering clock for better presentation)

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,35 @@
+$clock-size: 500px;
+
+.watch {
+  position: relative;
+  height: $clock-size;
+  width: $clock-size;
```
Ihor-Karpyn, line 6: mixin square

Ihor-Karpyn (approved):
LGTM👌

src/styles/blocks/watch.scss
illiarerun: Done, should've thought about that)

PR #630, opened 2021-07-14
Stop Watch Task

sasha-morozov:
- [DEMO LINK](https://sasha-morozov.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/hands-block.scss
+++ b/src/styles/blocks/hands-block.scss
@@ -1,0 +1,49 @@
+.hands {
+  @include clock-size($clock-size);
+  
+  position: relative;
+  background-color: #fffaf0;
+  border-radius: 50%;
+  border: 6px solid #dcdcdc;
+  
+  &__minute {
+    @extend %position_border-radius;
+    @include minute-size ($minute-height, $hands-width);
+
+    top: 25%;
+    background-color: #696969;
+    animation: move_minute 3600s steps(60, end) infinite;
+  }
+
+  &__second {
+    @extend %position_border-radius;
+    @include minute-size ($second-height, $hands-width);
+
+    background-color: #a9a9a9;
+    animation: move_second 60s steps(60, end) infinite;
+  }
+}
+
+@keyframes move_second {
+  0% {
+    transform: rotate(0deg);
+    transform-origin: bottom;
+  }
+
+  100% {
+    transform: rotate(360deg);
+    transform-origin: bottom;
+  }
+}
+
+@keyframes move_minute {
+  0% {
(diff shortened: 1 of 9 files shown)


ihor-jpeg (commented):
Great work. I have left a comment with suggested changes. Good luck!

Oh and yes, the hands of the clock are not positioned in the centre. You need to add transform: translateX() with an appropriate value to the hands and the animation itself.

src/styles/utilies/_mixins.scss
```diff
@@ -0,0 +1,14 @@
...
+
+@mixin minute-size ($height, $width) {
+  height: $height;
+  width: $width;
+}
+
+@mixin second-size ($height, $width) {
+  height: $height;
+  width: $width;
+}
```
ihor-jpeg, line 14: Here as well, two identical mixing. No need for that

sasha-morozov: They have only one thing in common - width. So, I decided to create separate mixins. However, I will try to adhere to your advice.

ihor-jpeg: They have the same properties. The values for them are passed by variables. so you can use one mixin in two different places, just pass different values.

sasha-morozov: > They have the same properties. The values for them are passed by variables. so you can use one mixin in two different places, just pass different values.

Thanks, I got you!:)

src/styles/blocks/hands-block.scss
```diff
@@ -0,0 +1,49 @@
...
+@keyframes move_minute {
+  0% {
+    transform: rotate(0deg);
+    transform-origin: bottom;
+  }
+
+  100% {
+    transform: rotate(360deg);
+    transform-origin: bottom;
+  }
```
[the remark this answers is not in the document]
VinnieJ-2k20: Agree with Ihor. Remove duplicate animation and use one twice.

src/styles/utilies/_extend.scss
```diff
@@ -0,0 +1,8 @@
+%position_border-radius {
+  position: absolute;
+  // left: 50%;
```
VinnieJ-2k20, line 3: Don't leave commented code.

VinnieJ-2k20 (approved):
Good job 👍 

Remember - don't repeat yourself. It is great if some code can be reused in multiple places in the project.

src/styles/utilies/_extend.scss
sasha-morozov: Thanks, will add changes accodringly to your advices!

PR #639, opened 2021-07-15
Hands-clock-home-work

AndriyNikorych:
- [DEMO LINK](https://AndriyNikorych.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/_frames.scss
+++ b/src/styles/_frames.scss
@@ -1,0 +1,9 @@
+@keyframes moveRound {
+  0% {
+    transform: rotate(180deg);
+  }
+
+  100% {
+    transform: rotate(540deg);
+  }
+}
--- a/src/styles/_mixin.scss
+++ b/src/styles/_mixin.scss
@@ -1,0 +1,14 @@
+@mixin positionClocks {
+  position: absolute;
+  left: $cirle-center-x;
+  top: $cirle-center-y;
+  height: $minute-hight;
+  width: $minute-width;
+  border-radius: 15px;
+}
+
+@mixin animationSet {
+  animation-timing-function: linear;
+  transform-origin: $cirle-center-x / 2 0;
+  animation-iteration-count: infinite;
+}
--- a/src/styles/clock.scss
+++ b/src/styles/clock.scss
@@ -1,0 +1,31 @@
+.clock {
+  position: relative;
+  top: 50%;
+  left: 50%;
+
+  &__sec {
+    @include positionClocks();
+    background-color: #c25d5d;
+
+    animation-name: moveRound;
+    animation-duration: 1s;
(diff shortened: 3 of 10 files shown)


src/styles/clock.scss
```diff
@@ -0,0 +1,31 @@
...
+  left: 50%;
+
+  &__sec {
+    @include positionClocks();
+    background-color: #c25d5d;
+
+    animation-name: moveRound;
+    animation-duration: 1s;
+    @include animationSet();
+  }
```
nastyyvolk, line 13: As a variant, you can include these two properties in your mixin, so it would take duration as a parameter
f.e. @include animation(1s)

src/styles/_mixin.scss
```diff
@@ -0,0 +1,14 @@
...
+  top: $cirle-center-y;
+  height: $minute-hight;
+  width: $minute-width;
+  border-radius: 15px;
+}
+
+@mixin animationSet {
+  animation-timing-function: linear;
+  transform-origin: $cirle-center-x / 2 0;
+  animation-iteration-count: infinite;
```
nastyyvolk, line 13: As for me, it's more convenient to use shorthand animation property

nastyyvolk (commented):
Great work! I've suggested a few changes you may want to think about.

src/styles/_mixin.scss
AndriyNikorych: Thanks. It is made for easier reading code, but you are right)

varseniuk: Both of you are correct

src/styles/clock.scss
varseniuk: Good point

src/styles/_frames.scss
```diff
@@ -0,0 +1,9 @@
+@keyframes moveRound {
+  0% {
+    transform: rotate(180deg);
+  }
+
+  100% {
+    transform: rotate(540deg);
```
varseniuk, line 7:
```suggestion
    transform: rotate(180deg);
  }

  100% {
    transform: rotate(540deg);
```
Why not 0 and 360 deg?

varseniuk (changes requested):
As task description says:

> Create a working stopwatch with minute and second hands

Now your watch is very fast. Let's make seconds and minutes

https://user-images.githubusercontent.com/66570142/126167743-9c3f9c96-b93d-4545-8f68-8e7dcb2b08c9.mov

src/styles/_frames.scss
AndriyNikorych: becouse they starts position is on 6:00, and i made 180 - 540 to start on 00:00

varseniuk (approved):

PR #643, opened 2021-07-15
watch done

karolina-nad:
- [DEMO LINK](https://karolina-nad.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,2 @@
+@import url(./normalize.scss);
+@import url(./watch.scss);
--- a/src/styles/watch.scss
+++ b/src/styles/watch.scss
@@ -1,0 +1,49 @@
+%hand {
+  position: absolute;
+  left: 49%;
+  width: 10px;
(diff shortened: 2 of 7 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,2 @@
+@import url(./normalize.scss);
+@import url(./watch.scss);
```
Serhii-Usenko, line 2: can use just "./........"
without url

Serhii-Usenko (commented):
great watch, please see my work

src/styles/main.scss
varseniuk: Good point

src/styles/watch.scss
```diff
@@ -0,0 +1,49 @@
+%hand {
+  position: absolute;
+  left: 49%;
```
varseniuk, line 3:
```suggestion
  left: 50%;
```
Why 49%?

varseniuk (approved):
Good job!

PR #658, opened 2021-07-19
watch done

yevhenpecherskyi:
https://yevhenpecherskyi.github.io/layout_stop-watch/


src/styles/utils/_mixin.scss
```diff
@@ -0,0 +1,5 @@
+@mixin pos {
```
varseniuk, line 1:
```suggestion
@mixin position {
```
`pos` is very strange name

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,42 @@
...
+  @keyframes round {
+    100% {
+      transform: rotate(360deg);
+    }
+  }
+
+  &::after {
+    position: absolute;
+    left: 195px;
+    top: 240px;
```
varseniuk, line 35: These numbers are not reliable. Better to use percentages

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,42 @@
...
+    width: 10px;
+    height: 150px;
+    border-radius: 5px;
+    background-color: #c25d5d;
+    transform-origin: bottom;
+    animation: round 3600s linear infinite;
+  }
+
+  &__arrow-sec {
+    position: absolute;
```
varseniuk, line 15: Why every element is absolute? Let the container be relative and all inner elements absolute

varseniuk (approved):
Good work. Try to not use 'magic numbers' in your future tasks and use calculation with `calc` function in css and percentages

PR #659, opened 2021-07-19
add clock

Andrii-Kalaida:
- [DEMO LINK](https://Andrii-Kalaida.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/body.scss
+++ b/src/styles/body.scss
@@ -1,0 +1,6 @@
+.body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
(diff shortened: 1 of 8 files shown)


src/styles/body.scss
```diff
@@ -0,0 +1,6 @@
+.body {
```
Ihor-Karpyn, line 1:
```suggestion
.page {
```

src/styles/clock.scss
```diff
@@ -0,0 +1,48 @@
+.clock {
+  position: relative;
+  width: $size-circle;
+  height: $size-circle;
```
Ihor-Karpyn, line 4: mixin square

src/styles/body.scss
Andrii-Kalaida: Ок. Виправлю!

PR #673, opened 2021-08-08
stop-watch Initial commit

GetLucky23:
Artem Motalov / GetLucky23

I did this task as version of clock (proper background-color and radius).
Arrow animations are smooth, like 'em that way. I can also make them go rapidly (not smooth).

Thanks for the review!


vkryvytskyy (approved):
great job!

PR #691, opened 2021-08-25
some message

ghost:
- [DEMO LINK](https://perronogrande.github.io/layout_stop-watch/)

COMMENT:
I didn't use a lot of css because I think this code is better this way ands css would only make it bigger and harder with no real purpose for that.


herkoss (approved):
Great!👍😊

PR #754, opened 2021-09-20
it`s done

DammneD:
- [DEMO LINK](https://dammned.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,72 @@
+$spot-radius: 31px;
+$second-width: 10px;
+$second-height: 250px;
+$minute-width: 14px;
+$minute-height: 200px;
+$second: 60s;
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,84 @@
...
+        $spot-radius
+      );
+    }
+  }
+}
+
+@keyframes spin {
+  from {
+    transform: rotate(0deg);
+  }
```
DionStrzh, line 80: You can set only "to"-step, 'cause start-position doesn't change. But your one is fine.

DionStrzh (approved):
Fancy watch. But be careful with setting border)

VinnieJ-2k20 (approved):
Good job 👍

PR #780, opened 2021-10-20
Develop stopwatch

darynadudar:
- [DEMO LINK](https://DudarDaryna.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,8 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
     >
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Stop watch</title>
@@ -12,10 +11,10 @@
 
   <body>
     <div class="stopwatch">
-      <article class="clock">
-        <div class="clock__minutes"></div>
-        <div class="clock__seconds"></div>
-        <div class="clock__center"></div>
+      <article class="watch stopwatch__watch">
+        <div class="watch__minutes"></div>
+        <div class="watch__seconds"></div>
+        <div class="watch__center"></div>
       </article>
     </div>
   </body>
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -3,21 +3,23 @@
 
 body {
   margin: 0;
-  background-color: #2d5074;
+  background-color: $dark-blue;
 }
 
 .stopwatch {
   height: 100vh;
+
+  &__watch {
+    width: $clock-size;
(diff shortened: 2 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,65 @@
...
+  background-color: #2d5074;
+}
+
+.stopwatch {
+  height: 100vh;
+}
+
+.clock {
+  width: $clock-width;
+  height: $clock-height;
```
dariia-romanova, line 15: If clock is always a circle, you can use one variable for width and height, same for the clock center

src/index.html
```diff
@@ -1,14 +1,22 @@
...
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+
+  <body>
+    <div class="stopwatch">
+      <article class="clock">
```
dariia-romanova, line 15: Is clock an element of a stopwatch? Maybe it should have class  stopwatch__clock

src/index.html
```diff
@@ -1,14 +1,22 @@
...
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
```
dariia-romanova, line 7: Write attribute from a new line
```suggestion
    <meta 
        name="viewport"
        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
    >
```

src/styles/main.scss
```diff
@@ -0,0 +1,65 @@
...
+    background-color: #2d5074;
+    border-radius: 4px;
+    animation: rotation 3600s linear infinite;
+
+    @extend %centring;
+  }
+
+  &__seconds {
+    width: 2px;
+    height: 220px;
```
dariia-romanova, line 37: You can bind height of clock's hands to a clock's size variable.

src/styles/main.scss
```diff
@@ -0,0 +1,65 @@
...
+
+.stopwatch {
+  height: 100vh;
+}
+
+.clock {
+  width: $clock-width;
+  height: $clock-height;
+
+  @extend %centring;
```
ihor-jpeg, line 17: your `.clock` is a BEM block. Do not give it positioning.
you already have a wrapper class `.stopwatch`. simply give your clock one more class, something like `stopwatch__clock`. this way your clock will have two classes: `clock` and `stopwatch__clock`. then move all the positioning to `stopwatch__clock`

src/styles/main.scss
ihor-jpeg: alternatively you can just skip setting the size and make your content dictate the clock's size

PR #850, opened 2021-11-25
Stop watch⏱

HrynykNataliia:
- [DEMO LINK](https://HrynykNataliia.github.io/layout_stop-watch/)


bcmp4p1 (approved):

Elena-brv (changes requested):
it should be seconds and minute arrows as on real watch (now it second and ms arrows)

src/styles/utils.scss/_extends.scss
```diff
@@ -0,0 +1,10 @@
+%flex-center {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+%positioning {
+  position: absolute;
+  bottom: 50%;
+}
```
Elena-brv, line 10: it's better to use mixin for this styles

Elena-brv (approved):

PR #858, opened 2021-11-26
Develop

sergeyglazgo:
- [DEMO LINK](https://sergeyglazgo.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -8,7 +8,6 @@
     transform: translate(-50%, -50%);
     width: 600px;
     height: 600px;
-    border-radius: 50%;
   }
 }
 
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,79 @@
+.page {
+  margin: 0;
+
+  &__watch {
+    position: absolute;
+    top: 50%;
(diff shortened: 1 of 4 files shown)


AlonaSnigur (approved):

src/styles/main.scss
```diff
@@ -0,0 +1,78 @@
...
+      from {
+        transform: rotate(270deg);
+        transform-origin: left;
+      }
+
+      to {
+        transform: rotate(630deg);
+        transform-origin: left;
+      }
+    }
```
Elena-brv, line 68: it is customary to write animation at the end of the file and not nest it in any selector

src/styles/main.scss
```diff
@@ -0,0 +1,78 @@
+.page {
```
Elena-brv, line 1: image

Elena-brv (changes requested):
you can create only one animation, and use different steps(animation-timing-function) for minutes and seconds)

src/styles/main.scss
sergeyglazgo: image

Elena-brv (approved):

PR #1010, opened 2022-02-23
task is done

ElizaChernysh:
https://elizachernysh.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,101 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+$watch-size: 510px;
+$circle-size: 20px;
+$minute-height: 100px;
+$arrow-width: 10px;
+$secunde-height: $minute-height + 150px;
+
+@mixin watch-circle {
+  height: $watch-size;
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,101 @@
...
+}
+
+%minute-arrow {
+  height: $minute-height;
+  width: $arrow-width;
+  background-color: rgb(255, 30, 0);
+  border-radius: 5px;
+}
+
+%secunda-arrow {
```
illiarerun, line 36:
```suggestion
%seconds-arrow {
```

src/styles/main.scss
```diff
@@ -0,0 +1,101 @@
...
+  @include watch-circle;
+  border: 4px solid rgba(78, 76, 76, 0.94);
+  background-color: rgb(192, 182, 182);
+
+  &::before {
+    content: "";
+    @extend %minute-arrow;
+    position: relative;
+    bottom: $minute-height / 2;
+    left: 15px;
```
illiarerun, line 77: why do you need these weird calculations and values? just use position: absolute with bottom 50% for all the elements inside a watch

PR #1036, opened 2022-03-15
Develop

DimaKotovich:
[DEMO](https://DimaKotovich.github.io/layout_stop-watch/)


AntonLevchuk (approved):
Greate job 🔥. I think would be better, if watch were in the middle of the page, but on my mind it's not critical

AlinaMatuschak (approved):
Perfect 🚀🚀🚀

PR #1059, opened 2022-03-28
stopwatch is created, all the styles are implemented

efdalyalcin:
- [DEMO LINK](https://efdalyalcin.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -34,6 +34,7 @@ body {
     left: ($stopwatch-size / 2) - ($hand-width / 2);
 
     animation-name: rotate;
+    animation-delay: 1s;
     animation-duration: $animation-minute-time * 60;
     animation-timing-function: linear;
 
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,86 @@
+$stopwatch-size: 500px;
+$hand-width: 10px;
+$small-hand-height: $stopwatch-size / 3;
+$center: $stopwatch-size / 2;
+
+//a minute is arranged for 6s since it is boring to watch
+$animation-minute-time: 6s;
+
+@mixin watchsize {
+  width: $stopwatch-size;
+  height: $stopwatch-size;
+}
+
+body {
+  margin: 0;
+}
+
+.page__container {
+  margin: 50px auto;
+}
+
+.container {
+  position: relative;
+
+  @include watchsize();
+
+  border-radius: 50%;
+
+  background-color: #ededed;
+
+  &__hand {
+    position: absolute;
+    top: $center - $small-hand-height;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,87 @@
...
+.page__container {
+  margin: 50px auto;
+}
+
+.container {
+  position: relative;
+
+  @include watchsize();
+
+  border-radius: 50%;
```
AndreiSuc, line 27: You can include the border-radius property into the `watchsize` mixin to save space.

src/styles/main.scss
```diff
@@ -0,0 +1,87 @@
...
+
+  &__hand {
+    position: absolute;
+    top: $center - $small-hand-height;
+    left: ($stopwatch-size / 2) - ($hand-width / 2);
+
+    animation-name: rotate;
+    animation-delay: 1s;
+    animation-duration: $animation-minute-time * 60;
+    animation-timing-function: linear;
```
AndreiSuc, line 39: Maybe you could combine these into one line to save space.

AndreiSuc (approved):
I have to give credit to your structured approach and good math. But I think you're placing too many blank lines, which makes reading your code a little confusing.

src/styles/main.scss
efdalyalcin: good catch

src/styles/main.scss
efdalyalcin: done

src/styles/main.scss
```diff
@@ -0,0 +1,79 @@
...
+$small-hand-height: $stopwatch-size / 3;
+$center: $stopwatch-size / 2;
+
+//a minute is arranged for 6s since it is boring to watch
+$animation-minute-time: 6s;
+
+@mixin watchSize {
+  width: $stopwatch-size;
+  height: $stopwatch-size;
+  border-radius: 50%;
```
ihor-jpeg, line 12: this does not seem reusable which is the idea behind mixins. also if your mixin does not take in any arguments, you should use a placeholder instead

src/styles/main.scss
```diff
@@ -0,0 +1,79 @@
...
+      height: ($stopwatch-size / 2);
+      width: ($hand-width / 2);
+
+      background-color: #da0037;
+    }
+
+    &--pin {
+      position: absolute;
+      top: $center - $hand-width;
+      left: $center - $hand-width;
```
ihor-jpeg, line 58: consider making a mixin for this. you use the same set of props in three places

ihor-jpeg (changes requested):
Almost there!

meforoma (approved):

PR #1060, opened 2022-03-28
Initial commit

AndreiSuc:
- [DEMO LINK](https://andreisuc.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,64 @@
+/* #region Normalize */
+body {
+  margin: 0;
+}
+/* #endregion */
+
+/*#region Variables */
+  $pin-size: 24px;
+/* #endregion */
+
+/* #region Mixins */
+
+@mixin position {
+  position: absolute;
+  left: 50%;
+  bottom: 50%;
+}
+
+@mixin pin_size($pin_size) {
+  width: $pin_size;
+  height: $pin_size;
+  border-radius: 50%;
+}
+
+/* #endregion */
+
+.minute {
+  @include position;
+  height: 175px;
+  width: 12px;
+  transform: translateX(-12.5%);
+  background-color: red;
+  animation: rotate 60s linear infinite 2s; //this has been sped up 60 times
+  transform-origin: bottom center;
+}
+
+.second {
+  @include position;
+  height: 300px;
+  width: 6px;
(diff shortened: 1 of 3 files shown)


59MrRobot (approved):
I don't see any problems with your code style.

src/styles/main.scss
```diff
@@ -0,0 +1,62 @@
...
+$pin-size: 24px;
+/* #endregion */
+
+/* #region Mixins */
+
+@mixin position {
+  position: absolute;
+  left: 50%;
+  bottom: 50%;
+}
```
ihor-jpeg, line 17: do not make mixins that do not take in any arguments. use placeholder instead

src/styles/main.scss
```diff
@@ -0,0 +1,62 @@
...
+
+@mixin pin_size($pin_size) {
```
ihor-jpeg, line 19: i think circle would be a more descriptive name of what it is

src/styles/main.scss
```diff
@@ -0,0 +1,62 @@
...
+  background: linear-gradient(rgb(150, 150, 150),#000);
+  animation: rotate 60s linear infinite 2s;
+  transform-origin: bottom center;
+}
+
+.pin {
+  @include position;
+  @include pin_size($pin-size);
+  background-color: black;
+  transform: translate3D(-25%, 50%, 0);
```
ihor-jpeg, line 51: you are not utilizing the Z axis. simple translate would do

ihor-jpeg (changes requested):
Well done! Almost there!

ihor-jpeg (commented):
Also, I'm very confused by that apple logo 🤣

andreisucman:
The logo was supposed to be on the background but then I changed my mind)))

src/styles/main.scss
andreisucman: Fixed

src/styles/main.scss
andreisucman: Ok. Changed to circle.

src/styles/main.scss
andreisucman: Ok. Added an argument to the mixin.

meforoma (approved):

PR #1061, opened 2022-03-28
completed stopwatch task

59MrRobot:
- [DEMO LINK](https://59MrRobot.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -23,7 +23,6 @@
     }
 
     &--min {
-      position: absolute;
       top: 0;
       left: $center - ($hand-width / 4);
 
@@ -40,7 +39,6 @@
     }
 
     &--pin {
-      position: absolute;
       top: $center - $hand-width;
       left: $center - $hand-width;
 


AndreiSuc (approved):
Good job. Satisfies the checklist and has ideally calculated positioning.

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin watchSize($size) {
```
ihor-jpeg, line 1: it seems to me square would be a more informative name of what it does

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,5 @@
+.page {
+  &__clock {
+    margin: 60px auto;
+  }
+}
```
ihor-jpeg, line 5: i think a clock can be a block

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,52 @@
...
+      background-color: black;
+    }
+
+    &--pin {
+      position: absolute;
+      top: $center - $hand-width;
+      left: $center - $hand-width;
+
+      width: $hand-width * 2;
+      height: $hand-width * 2;
```
ihor-jpeg, line 44: reuse your mixin

src/index.html
```diff
@@ -1,14 +1,18 @@
...
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body class="page">
+    <div class="clock page__clock">
+      <div class="clock__hand"></div>
+      <div class="clock__hand clock__hand--big"></div>
+      <div class="clock__hand clock__hand--pin"></div>
```
ihor-jpeg, line 15: this is not clock hand. in the --pin selector you reset the majority of styles anyway

ihor-jpeg (changes requested):
Well done! Take a look at the comments

src/styles/blocks/clock.scss
ihor-jpeg: maybe throw in the border radius as well, and name it circle

src/styles/blocks/page.scss
59MrRobot: clock is a block, I did this so that I could set the block margins using mixes

ihor-jpeg: gotchu

ihor-jpeg (approved):
Nooice!

meforoma (approved):

PR #1062, opened 2022-03-29
Clock

huberciak:
- [DEMO LINK](https://huberciak.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,69 @@
+body {
+  margin: 0;
+}
+
+$size: 600px;
+
+.stopwatch {
+  position: relative;
+  background-color: rgb(236, 232, 232);
+  width: $size;
+  height: $size;
+  margin: 50px auto;
+  border-radius: 50%;
+
+    &__dot {
+      position: absolute;
+      background-color: rgb(236, 12, 12);
+      width: 0.1 * $size;
+      height: 0.1 * $size;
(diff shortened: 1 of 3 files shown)


AndreiSuc (approved):
You can turn it into an NFT bro.

huberciak:
> You can turn it into an NFT bro.

Time to launch Bored Stopwatch set.

src/styles/main.scss
```diff
@@ -0,0 +1,69 @@
...
+  margin: 0;
+}
+
+$size: 600px;
+
+.stopwatch {
+  position: relative;
+  background-color: rgb(236, 232, 232);
+  width: $size;
+  height: $size;
```
ihor-jpeg, line 11: consider making mixin along with the border radius

src/styles/main.scss
```diff
@@ -0,0 +1,69 @@
...
+      top: ($size - 0.025 * $size) / 2;
+      left: $size / 2;
+      z-index: 0;
+
+      transform-origin: 0;
+
+      animation-name: rotate;
+      animation-duration: 6s;
+      animation-timing-function: linear;
+      animation-iteration-count: 60;
```
ihor-jpeg, line 56: one hand will stop but the other one wont

ihor-jpeg (changes requested):
Checklist p1
- The starting position of the hands should be at 0:00

PR #1063, opened 2022-03-29
initial commit

amitpatiljc:
- [DEMO LINK](https://amitpatiljc.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,72 @@
+$watch-size: 600px;
+$watch-radius: $watch-size/2;
+$watch-minute-height: $watch-radius / 2;
+$watch-second-height: $watch-radius;
+$watch-minute-width: 6px;
+$watch-second-width: $watch-minute-width*2;
+$pin-size: 40px;
+$stop-watch-time: 60s;
+
+@mixin circle ($size) {
+  height: $size;
+  width: $size;
+  border-radius: 50%;
+}
+
+body {
+  margin: 0;
+}
+
+.watch {
+  position: relative;
+  @include circle($watch-size);
+  background-color: rgb(156, 171, 184);
+}
+
+.page__watch {
+  margin: 0 auto;
+}
+
+.watch__hand {
+  position: absolute;
+
+}
+
+.watch__hand--min {
+  height: $watch-minute-height;
+  width: $watch-minute-width;
+  top: $watch-radius - $watch-minute-height;
+  left: $watch-radius - $watch-minute-width;
+  background-color: blue;
+  transform-origin: bottom right;
+  animation: rotate $stop-watch-time*60 linear infinite;
+}
+
+.watch__hand--sec {
+  height: $watch-radius;
+  width: $watch-second-width;
+  background-color: red;
+  top: 0;
+  left: $watch-radius - $watch-second-width;
+  transform-origin: bottom right;
+  animation: rotate $stop-watch-time linear infinite;
+
+}
+
+.watch__hand--pin {
+  @include circle($pin-size);
+  top: 50%;
+  left: 50%;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,72 @@
...
+  transform-origin: bottom right;
+  animation: rotate $stop-watch-time linear infinite;
+
+}
+
+.watch__hand--pin {
+  @include circle($pin-size);
+  top: 50%;
+  left: 50%;
+  transform-origin: 0 0;
```
AndreiSuc, line 60: This `transform-origin: 0 0` makes your handles rotate around their bottom left corners. Probably `transform-origin: bottom center` will do a better job and they will be more aligned.

src/styles/main.scss
```diff
@@ -0,0 +1,72 @@
...
+  animation: rotate $stop-watch-time*60 linear infinite;
+}
+
+.watch__hand--sec {
+  height: $watch-radius;
+  width: $watch-second-width;
+  background-color: red;
+  top: 0;
+  left: $watch-radius - $watch-second-width;
+  transform-origin: bottom right;
```
AndreiSuc, line 51: `transform-origin: bottom right ` makes the figure rotate around its lower right corner. That's why the rotation is a little asymmetric. If you try `transform-origin: bottom center; `it should work better.

AndreiSuc, line 51: `transform-origin: bottom right ` makes the figure rotate around its lower right corner. That's why the rotation is a little asymmetric. If you try `transform-origin: bottom center; `it should work better.

src/styles/main.scss
```diff
@@ -0,0 +1,72 @@
...
+  animation: rotate $stop-watch-time*60 linear infinite;
+}
+
+.watch__hand--sec {
+  height: $watch-radius;
+  width: $watch-second-width;
+  background-color: red;
+  top: 0;
+  left: $watch-radius - $watch-second-width;
```
AndreiSuc, line 50: If you want to position your handle in the center you need to subtract half of its thickness, not the full thickness. So   left: `$watch-radius - ($watch-second-width / 2);` will look better.

AndreiSuc, line 50: If you want to position your handle in the center you need to subtract half of its thickness, not the full thickness. So   left: `$watch-radius - ($watch-second-width / 2);` will look better.

src/styles/main.scss
AndreiSuc, line 60: Since the pin is not rotating you probably don't need `transform-origin` here.

AndreiSuc, line 60: Since the pin is not rotating you probably don't need `transform-origin` here.

AndreiSuc (changes requested):
Hi Amit. I noticed some asymmetry in your clock and inspected it clock in chrome dev tools. Please have a look at my suggestions.

Apart from the suggestions, I like that you tied everything to variables, something I was lazy to do. 👍

src/styles/main.scss
amitpatiljc: Right! its looking great!

AndreiSuc (approved):
Looks great.

src/styles/main.scss
```diff
@@ -0,0 +1,75 @@
...
+  transform-origin: bottom center;
+  animation: rotate $stop-watch-time linear infinite;
+
+}
+
+.watch__hand--pin {
+  @include circle($pin-size);
+  top: 50%;
+  left: 50%;
+  // transform-origin: 0 0;
```
ihor-jpeg, line 63: no commented code

ihor-jpeg (commented):
Nicely done!

meforoma (approved):

PR #1064, opened 2022-03-29
A Rolex clock has been designed

leny-jeanbaptiste:
[DEMO LINK](https://leny-jeanbaptiste.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,21 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,14 +1,21 @@
...
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
```
amitpatiljc, line 12: you can use 'hidden' attribute here.

src/styles/main.scss
```diff
@@ -0,0 +1,57 @@
...
+    border-radius: 50%;
+    top: 40%;
+    left: 49%;
+    background-color: black;
+  }
+
+  &__hour,
+  &__minute,
+  &__second {
+    transform-origin: 0 100%;
```
amitpatiljc, line 49: I think there is some problem with the symmetry of origin of all hands . If you check in dev tools you can see that all hands are not in sync as far as their point around which they revolve.

src/styles/main.scss
```diff
@@ -0,0 +1,57 @@
+.clock {
+  height: 340px;
```
amitpatiljc, line 2: This is optional task but when you get time try to work with variables for this task and try to formulate some values. 
Avoid hard code wherever you can

src/styles/main.scss
ihor-jpeg: agree

ihor-jpeg (changes requested):
It seems like the clock hands are not in the middle

PR #1065, opened 2022-03-30
Develop

MabuzaM:
- [DEMO LINK](https://MabuzaM.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,109 @@
...
+    3600s,
+    steps(1, end),
+    infinite,
+    bottom
+  );
+
+}
+
+.watch__hand--second {
+  position: absolute;
```
59MrRobot, line 51: Because this a watch__hand modifier, position will be inherited from watch__hand therefore it is redundant here.

src/styles/main.scss
```diff
@@ -0,0 +1,109 @@
...
+    60s,
+    steps(60, end),
+    infinite,
+    bottom
+  );
+
+}
+
+.watch__hand--minute {
+  position: absolute;
```
59MrRobot, line 70: Because this a watch__hand modifier, position will be inherited from watch__hand therefore it is redundant here.

src/styles/main.scss
```diff
@@ -0,0 +1,109 @@
...
+  background-color: #000;
+  border-radius: 50%;
+
+  @include animation-props(
+    rotate,
+    3600s,
+    steps(1, end),
+    infinite,
+    bottom
+  );
```
59MrRobot, line 46: I believe these styles were supposed to be in your watch__hand--hour class.

src/styles/main.scss
```diff
@@ -0,0 +1,109 @@
...
+
+.watch {
+  width: 500px;
+  height: 500px;
+  border-radius: 50%;
+  background: rgb(211, 141, 11);
+  border: 12px double #333;
+}
+
+.watch__hand {
```
59MrRobot, line 31: i would recommend nesting elements in their blocks using `&`

59MrRobot (changes requested):
Good job, just a couple of suggestions.

src/styles/main.scss
MabuzaM: Okay noted

src/styles/main.scss
MabuzaM: Okay thanx noted

ihor-jpeg (changes requested):
Well done! It does not seem like the clock hands are located in the middle

src/styles/main.scss
MabuzaM: > i would recommend nesting elements in their blocks using `&`

Hi. Have made changes. Please have a look.

MabuzaM:
> Well done! It does not seem like the clock hands are located in the middle

Hi. Have made changes. Please have a look.

meforoma (approved):

PR #1075, opened 2022-04-03
Stop-watch

LukaszGryziewicz:
I completed the task with grid, since it seemed to be the most efficient, I hope no one will mind since this is an optional task 😄 

P.S. I forgot to add variables to git in the first commit, so they are located in the second one.

- [DEMO LINK](https://LukaszGryziewicz.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,45 @@
+.clock {
+  display: grid;
+  justify-items: center;
+  align-items: center;
+  height: 100vh;
+
+  //this gives the same position to all children
+  & * {
+    grid-column: 1;
+    grid-row: 1;
+  }
+
+  &__pin {
+    z-index: 1;
+    @include circle(25px);
+    background-color: $pale-blue;
+  }
+
+  &__hand-minute {
+    @include rotateClockwise(60s);
+    width: 15px;
+    height: $hand-minute-height;
+    border-radius: 8px;
+    margin-bottom: $hand-minute-height;
+    background-color: $pale-blue;
+  }
+
+  &__hand-second {
+    @include rotateClockwise(1s);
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,11 @@
+@import "utils/variables";
+@import "utils/mixins";
+@import "blocks/clock";
+
+* {
+  box-sizing: border-box;
```
andreisucman, line 6: When I remove this `box-sizing` property nothing changes, so it's probably redundant here. But I wouldn't call it a mistake.

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,45 @@
...
+  display: grid;
+  justify-items: center;
+  align-items: center;
+  height: 100vh;
+
+  //this gives the same position to all children
+  & * {
+    grid-column: 1;
+    grid-row: 1;
+  }
```
andreisucman, line 11: Great idea,

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,45 @@
...
+
+  @keyframes rotate {
+    0% {
+      transform: rotate(0deg);
+    }
+    100% {
+      transform: rotate(360deg);
+    }
+  }
```
andreisucman, line 44: Maybe animations also deserve a separate file, but I'm not sure.

andreisucman (approved):
This watch is fire.

src/styles/blocks/clock.scss
ihor-jpeg, line 11: does every parent also has display grid? because children only means direct children

ihor-jpeg (approved):
Well done!

PR #1078, opened 2022-04-05
the clock is ticking

ZUBOV-ILLIA:
- [DEMO LINK](https://ZUBOV-ILLIA.github.io/layout_stop-watch/)


ZUBOV-ILLIA:
The main idea of using mixins and variables here is that changing one clock size value affects all other internal sizes!

EnjSadman (approved):

Anna-Tkachenko (approved):
Well Done! ✔️

PR #1098, opened 2022-04-11
Watches first re

ghost:
- [DEMO LINK](https://Oleksij-Kozakovskyj.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,8 +1,4 @@
 .clock-face {
-  box-sizing: border-box;
-  border: solid;
-  border-radius: 50%;
-  border-color: black;
   height: 500px;
   width: 500px;
   position: relative;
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -1,14 +1,19 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="arrow-min"></div>
+    <div class="arrow-sec"></div>
+    <div class="clock-face">
+      <div class="clock-face__fix"></div>
+    </div>
```
ZUBOV-ILLIA, line 16: `<div class="clock-face">` unnecessary here.
you can leave only `<div class="clock-face__fix"></div>`

ZUBOV-ILLIA (changes requested):
Something wrong with positions, because arrows not in center.

src/index.html
Denchepornyuk, line 16: I believe you can use one div instead of two

Denchepornyuk (approved):

PR #1152, opened 2022-05-17
Clock done

dimabelotskyi:
- [DEMO LINK](https://dimabelotskyi.github.io/layout_stop-watch/)


ant2nta (approved):

mykola-koval (approved):
nice job! 
but in real life, the minute hand is usually shorter than the second hand and it is also moving a little step every second, not once a minute.
not a problem, just a little unusual behavior to mention :)

PR #1222, opened 2022-06-08
The clock is ticking

LordAsmodey:
- [DEMO LINK](https://LordAsmodey.github.io/layout_stop-watch/)


VETALEK (approved):
GJ 🚀 

you may use [`animation-timing-function: steps`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-timing-function) for more ticking animation 😉

PR #1230, opened 2022-06-10
Stopwatch created

ElinaBahirova:
(https://elinabahirova.github.io/layout_stop-watch/)


Bohdan-Maksymiuk (changes requested):
Good job, but make your clock smaller (it should fit the viewport without scrolling)

PR #1234, opened 2022-06-11
Stopwatch

vitalii-labunskyi:
- [DEMO LINK](https://vitalii-labunskyi.github.io/layout_stop-watch/)


aliceblunt8 (changes requested):
the clock should behave like the REAL clock

vitalii-labunskyi:
> the clock should behave like the REAL clock

Oh, I thought that it should be a stopwatch and not a watch ))

aliceblunt8 (approved):

PR #1250, opened 2022-06-14
clocks was created

DashaPoliazz:
- [DEMO LINK](https://DashaPoliazz.github.io/layout_stop-watch/)


src/styles/components/clock.scss
```diff
@@ -0,0 +1,63 @@
...
+
+  &__minute {
+    position: absolute;
+    top: 30%;
+    left: 50%;
+    height: 100px;
+    width: 5px;
+    background-color: $hand-color;
+    border-radius: 50%;
+    transform-origin: center bottom;
```
butenko108, line 50: Why bottom?
transform-origin: center bottom;

Why not?
transform-origin: center left;

image

src/styles/main.scss
```diff
@@ -0,0 +1,7 @@
+@import "./utils/variables";
+@import "./utils/mixins";
+@import "./components/clock";
+
+* {
+  box-sizing: border-box;
+}
```
butenko108, line 7: Main.scss should only have @import.

PR #1255, opened 2022-06-14
Develop

frontend112:


ihor-jpeg (changes requested):
Deploy your project and make sure that the demo link is working and is present in the PR description

src/styles/utils/_count.scss
```diff
@@ -0,0 +1,7 @@
+@keyframes count {
+  @for $i from 0 through 12 {
+    #{$i * 100% / 12} {
+      transform: rotate(calc((-90deg + 30 * #{$i}deg)));
+    }
+  }
```
meforoma, line 6: it's a bit overwhelming
wouldn't be cleaner to define start and end %?

```suggestion
  0% {
    transform: rotate(0);
  }

  100% {
    transform: rotate(360deg);
  }
}
```

meforoma (changes requested):
the amination seems more complicated than it can be

meforoma (approved):
👍

PR #1289, opened 2022-06-17
420 o'clock

Vova-Balyshev:
[Acid clock](https://Vova-Balyshev.github.io/layout_stop-watch/)
Its always starts at 4:20. You can press a button to refresh your time ^^
Enjoy =)

### DIFF
--- a/src/styles/utils/_mixins.scss
+++ b/src/styles/utils/_mixins.scss
@@ -1,0 +1,4 @@
+@mixin circle($size) {
+  height: $size;
+  width: $size;
+}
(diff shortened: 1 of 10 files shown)


src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin circle($size) {
+  height: $size;
+  width: $size;
+}
```
VETALEK, line 4: I don't believe you, it is a square 😁
use `border-radius` here 😉

VETALEK (changes requested):
Looks good 🤩 

mb better to move the button to the center 😉

image

VETALEK (approved):
Like 👍

PR #1300, opened 2022-06-21
Done

Ostap4yk:
[DEMO LINK](https://Ostap4yk.github.io/layout_stop-watch/)


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,32 @@
+.stopwatch {
+  position: relative;
+
+  display: flex;
+  width: $container-size;
+  height: $container-size;
+  margin: auto;
```
BudnikOleksii, line 7: You can't use margins to BEM-block, 8-th from checklist:
[BEM & STYLES] - Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

src/styles/utils/_extends.scss
```diff
@@ -0,0 +1,13 @@
...
+  left: $hand-x-pos;
+
+  width: $hand-width;
+
+  border-radius: 6px;
+
+  transform-origin: bottom;
+  animation-name: time;
+  animation-timing-function: steps(60, end);
+  animation-iteration-count: infinite;
```
BudnikOleksii, line 12: You can write it in short form, only for property animation😉

BudnikOleksii (changes requested):
Nice 👍
Fix margins

liliia-livandovskaya (approved):
GJ! Pay attention to the comments

PR #1309, opened 2022-06-26
working

illia-kyselov:


src/styles/main.scss
```diff
@@ -0,0 +1,112 @@
...
+    position: absolute;
+
+    .minutes {
+      display: flex;
+      justify-content: center;
+      position: relative;
+      width: 130px;
+      height: 130px;
+      border-radius: 50%;
+      animation: minutes 720s linear infinite;
```
Bohdan-Maksymiuk, line 61:
```suggestion
      animation: minutes 3600s linear infinite;
```
why 720?
now after 1 turn of second arrow it will pass near 5 minutes...

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1339, opened 2022-07-15
layout_stop-watch task done

EhorSaprykin:


src/styles/utils/_vars.scss
```diff
@@ -0,0 +1,2 @@
+$minute-circle: 60s;
+$hour-circle: (60* 60 * 24) * 1s;
```
valerii-derkach, line 2: Your minute arrow moves like second arrow, it is ok, but your hour arrow moves like hour arrow, so you do not have a minute arrow.
you have second arrow and hour arrow, make your hour arrow to move as minute arrow 60 * 60

valerii-derkach (changes requested):

src/styles/utils/_vars.scss
EhorSaprykin: @valerii-derkach done

ArtemTopolskyi (approved):

valerii-derkach (approved):

PR #1361, opened 2022-07-20
CSS Watch Added

kir8mir:
- [DEMO LINK](https://kir8mir.github.io/layout_stop-watch/)


src/styles/blocks/card.scss
```diff
@@ -0,0 +1,221 @@
...
+    flex-wrap: wrap;
+    justify-content: center;
+    gap: 25px;
+
+    .card {
+      position: relative;
+
+      display: flex;
+      flex-direction: column;
+      align-items: center;
```
AndriiRut, line 30: use mixins for styles which you used more than 2 times

AndriiRut (changes requested):
The card and watch can be separate blocks, so they should be in different files.
Good job!
MJ Approved 👍🏻

AndriiRut (approved):
🔥

PR #1364, opened 2022-07-21
solution

ostUp:
- [DEMO LINK](https://ostUp.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,9 @@
+* {
```
oksana-kyryienko, line 1: Don't use margin and padding styles with * selector

danylo-onishchuk (approved):

PR #1367, opened 2022-07-21
watch was added

Olena-Shevchenko:
- [DEMO LINK](https://Olena-Shevchenko.github.io/layout_stop-watch/)


ArturErmolenko (approved):
In this case ,all right with BEM 👍

PR #1372, opened 2022-07-21
watch

Maxim-Gumeniuk:
[DEMO LINK](https://Maxim Gumeniuk.github.io/layout_stop-watch/)


AndreyJigijalkin (approved):
Generally it's nice project, but in my opinion, it would be good to create variables for watch size and make all other parts (arrows, border-thickness) dependant from watch size, so it could look similar either it 500px or 300px:)

Maxim-Gumeniuk:
> Generally it's nice project, but in my opinion, it would be good to create variables for watch size and make all other parts (arrows, border-thickness) dependant from watch size, so it could look similar either it 500px or 300px:)

it was good advice.I will use them in next project.Thank you!)

danylo-onishchuk (approved):

PR #1373, opened 2022-07-21
Solution clock

oksana-kyryienko:
- [DEMO LINK](https://oksana-kyryienko.github.io/layout_stop-watch/)


src/styles/blocks/reset.scss
```diff
@@ -0,0 +1,5 @@
+* {
+  margin: 0;
+  padding: 0;
```
ostUp, line 3: Don't use margin and padding styles with * selector

src/styles/main.scss
```diff
@@ -0,0 +1,7 @@
+@import "./utils/variables";
+@import "./blocks/clock.scss";
+@import "./blocks/reset.scss";
+
+body {
+  background-color: lightblue;
+}
```
ostUp, line 7: I think you can move background-color to clockflex class🤔

PR #1381, opened 2022-07-22
stopwatch

adrqq:
[DEMO LINK](https://adrqq.github.io/layout_stop-watch/)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,66 @@
+
+.clock {
+  position: relative;
+  height: $clock-size;
+  width: $clock-size;
+  margin: auto;
```
ostUp, line 6: You shouldn't use margin for the BEM block

danylo-onishchuk (approved):

PR #1424, opened 2022-08-02
the clock was added

smyrennia:
https://smyrennia.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,125 @@
+.clock {
+  position: absolute;
+  height: $clock-size;
+  width: $clock-size;
+  background-color: #fff;
+  border-radius: 100%;
+  border: 10px solid #34626c;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  margin: auto;
+
+  &__top {
+    position: absolute;
+    width: 3px;
+    height: 8px;
+    background-color: $repeated-color;
+    left: 0;
+    right: 0;
+    margin: 0 auto;
+  }
+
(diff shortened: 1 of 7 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,125 @@
...
+    height: 100%;
+    position: absolute;
+    left: 0;
+    right: 0;
+    margin: 0 auto;
+    animation: move 60s steps(60) infinite;
+
+    &::before {
+      content: "";
+      background-color: $repeated-color;
```
marina-tilniak, line 118: let's move common style to common class

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,125 @@
...
+    right: 0;
+    margin: auto;
+  }
+
+  &__bottom {
+    position: absolute;
+    width: 3px;
+    height: 8px;
+    background-color: $repeated-color;
+    left: 0;
```
marina-tilniak, line 40: let's move common style to common class

marina-tilniak (changes requested):
Nice styles!

src/styles/blocks/clock.scss
smyrennia: could you explain more, please? because I don't really understand

marina-tilniak: if you have same styles for 2 classes, you can create one class hand for example and then just create modifier for minute hand with different styles. If you have other question you can ask me in your channel, just tag me

Bohdan-Maksymiuk (approved):

PR #1425, opened 2022-08-02
stop-watch added

KseniyaGonchar:
https://KseniyaGonchar.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,54 @@
+@keyframes rotate {
+  from {
+    transform: rotate(0);
+  }
+  to {
+    transform: rotate(1turn);
+  }
+}
+
+.container {
+  display: flex;
+  height: 100vh;
+}
+
+.stop-watch {
+  margin: auto;
+  width: 600px;
+  height: 600px;
+  border: 1px solid rgb(87, 134, 216);
+  border-radius: 50%;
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,54 @@
...
+    animation: rotate linear 60s infinite;
+  }
+
+  &__hand--circle {
+    width: 20px;
+    height: 20px;
+    border-radius: 50%;
+    background-color: blue;
+    left: calc(50% - 10px);
+    top: calc(50% - 10px);
```
Anna-Tkachenko, line 52: Use variables for main values
For example, here will be better to create variables for hands sizes, and it will be clear to understand why you set -10 px when calculating the position of hand
So create main variables for hand size, and use them for calculating the hand position

Anna-Tkachenko (changes requested):

src/styles/main.scss
```diff
@@ -0,0 +1,58 @@
+$width-hand: 8px;
+$width-circle: 20px;
+$height-circle: 20px;
+
+@keyframes rotate {
```
alexandra-protyanova, line 5: move variables, keyframes to separate files

alexandra-protyanova (approved):

PR #1427, opened 2022-08-03
03.08

HelenaVer:
https://helenaver.github.io/layout_stop-watch/


src/styles/blocks/normalize.scss
```diff
@@ -0,0 +1,3 @@
+* {
+  margin: 0;
+}
```
Anna-Tkachenko, line 3: Will be better not to set margin: 0 to all elements
Just specify needed elements, for which you need to reset margins

Anna-Tkachenko (approved):
Good job!

PR #1440, opened 2022-08-09
Stop-watch was created

SergIats:


SergIats:
[DEMO LINK](https://SergIats.github.io/layout_stop-watch/)

src/styles/blocs/stopwatch.scss
```diff
@@ -0,0 +1,87 @@
+.stopwatch {
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  &__screen {
+    position: relative;
+    display: flex;
```
alexandra-protyanova, line 9: create more variables for calculations

alexandra-protyanova (changes requested):

src/styles/blocs/stopwatch.scss
SergIats: Alexandra, hi! Thanks for recommendation. I have tried to do it, but on the first step for some reason I saw service message "C:\Users\iatse\projects\layout_stop-watch\src\styles\blocs\stopwatch.scss:25:25: Undefined variable: "$hour-step".
1. I tried to change order of import-link in main.scss
2. I tried to use  @use instead @import
3. I tried different format of file name - with "_" in begin and without it.

With mixins and extends I have the same problem. Can you help me to understand what is the problem?

SergIats:
Hi! I added some variables, mixins and placeholders, but I have added it directly in SCSS file where they must work. I know that is not the best way, but neither @import nor @use hadn't worked this time, and I don't know why.

src/styles/blocs/stopwatch.scss
SergIats: Alexandra, hi! I had had some problems with @import, but finally I have created variables and mixins. Check it, please!

marina-tilniak (approved):

PR #1441, opened 2022-08-10
made watch

Denis-Vasileha:
[DEMO LINK](https://Denis-Vasileha.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,89 @@
+body {
+  margin: 0;
+  padding: 0;
+  height: 100vh;
+
+  display: flex;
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,89 @@
...
+}
+
+@keyframes sec {
+  0% {
+    transform: rotate(0deg);
+  }
+
+  100% {
+    transform: rotate(360deg);
+  }
```
alexandra-protyanova, line 88: they are the same U can use 1 of them & nothing will change

alexandra-protyanova (approved):

PR #1448, opened 2022-08-13
Added watch in index.html

dmytro-klepikov:
https://dmytro-klepikov.github.io/layout_stop-watch


src/styles/blocks/_watch.scss
```diff
@@ -0,0 +1,42 @@
...
+    @extend %arrow-position;
+    @extend %move-arrow;
+    width: $small-width;
+    height: $small-height;
+    background-color: rgb(10, 69, 91);
+    border: 2px solid rgb(201, 201, 201);
+    border-radius: 5px;
+    animation-duration: 3600s;
+  }
+
```
marina-tilniak, line 34: let's move common styles to common class

marina-tilniak (changes requested):
Nice styles 👍

marina-tilniak (changes requested):
why do you need 2 files with styles (watch and _watch)?

Virass (changes requested):
One of the files watch, _watch need to be deleted.

PR #1470, opened 2022-08-20
First Commit

rstarovoit:


package.json
```diff
@@ -23,7 +23,7 @@
     "@mate-academy/bemlint": "^0.1.1",
```
grafit-off, line 23: Don't add such files to the PR.

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,66 @@
...
+      height: $seconds-hand-length;
+
+      animation-duration: 60s;
+    }
+
+    &--minutes {
+      width: $minutes-hand-thickness;
+      background-color: black;
+      height: $minutes-hand-length;
+
```
grafit-off, line 52: You should add linear animation-timing-function, because time looks strange when time starts faster and ends slower (visually on the stopwatch)

grafit-off (approved):
Good job! Keep it up! :)

Anna-Tkachenko (approved):
Awesome! ⭐️

PR #1480, opened 2022-08-22
created watch

ArtemHulkoDev:
- [DEMO LINK](https://ArtemHulkoDev.github.io/layout_stop-watch/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # Stopwatch task
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_stop-watch/)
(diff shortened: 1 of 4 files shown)


src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,26 @@
...
+  top: 5px;
+  left: 15px;
+
+  border-radius: 8px;
+
+  transform-origin: left;
+
+  animation-name: circle;
+  animation-iteration-count: infinite;
+  animation-timing-function: linear;
```
Vladosik1337, line 25: It seems to me that it would be better to write everything in one line, but this is not an error i think
`animation: circle infinite linear`

Vladosik1337 (approved):
Nice work)

MaksimusCRS (approved):
GJ

PR #1481, opened 2022-08-22
Create a page with clock

Vladosik1337:
- [DEMO LINK](https://Vladosik1337.github.io/layout_stop-watch/)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,55 @@
...
+  border-radius: 50%;
+
+  background: url("../img/5-55539_transparent-broken-clock-clipart-transparent-background-clock-face.png") #fff no-repeat center;
+  background-size: cover;
+
+}
+
+.clock::before {
+  content: "";
+
```
ArtemHulkoDev, line 17: only recommendation to use nesting during elements styling

ArtemHulkoDev (approved):
everything is fine, GJ

src/styles/blocks/clock.scss
OleksiiBlezniuk: Agree with @ArtemHulkoDev , use sass nesting

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,25 @@
+@mixin dot-size {
+  height: 20px;
+  width: 20px;
+}
+
+@mixin clock-size {
+  height: 500px;
+  width: 500px;
+}
```
OleksiiBlezniuk, line 9: It's a same logic, create one mixin and pass variable

OleksiiBlezniuk: Also in both cases you setting border-radius for elements, so let's set it here to have reusable cirlce

OleksiiBlezniuk (changes requested):
Awesome UI! 🔥 

Lets create variables for main values (sizes, duration, etc.)

PR #1483, opened 2022-08-22
implemented stopwatch

ihorandrianov:
- [DEMO LINK](https://ihorandrianov.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/_watch.scss
+++ b/src/styles/blocks/_watch.scss
@@ -3,7 +3,7 @@
   justify-content: center;
   align-items: center;
   position: relative;
-  margin-top: 50px;
+
 
   &__circle {
     @include circle(600px);
@@ -11,9 +11,9 @@
     background: linear-gradient(-45deg, #ee7752, #e73c7e, #23a6d5, #23d5ab);
     background-size: 400% 400%;
 
-    animation: blobs 15s ease-in-out infinite alternate,
-      bg-gradient 15s ease infinite;
-    transition: 2s;
+    animation: blobs $clockface-duration ease-in-out infinite alternate,
+      bg-gradient $clockface-duration ease infinite;
+    transition: scale 2s;
 
     &:hover {
       scale: 80%;
--- a/src/styles/blocks/_watch.scss
+++ b/src/styles/blocks/_watch.scss
@@ -1,0 +1,87 @@
+.watch {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  position: relative;
+  margin-top: 50px;
+
+  &__circle {
+    @include circle(600px);
+
+    background: linear-gradient(-45deg, #ee7752, #e73c7e, #23a6d5, #23d5ab);
+    background-size: 400% 400%;
+
+    animation: blobs 15s ease-in-out infinite alternate,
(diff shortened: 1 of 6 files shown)


src/styles/blocks/_watch.scss
```diff
@@ -0,0 +1,87 @@
+.watch {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  position: relative;
+  margin-top: 50px;
```
grafit-off, line 6: It's a BEM block, you can't set margins on it

src/styles/blocks/_watch.scss
```diff
@@ -0,0 +1,87 @@
...
+  &__center-point {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    @include circle(20px);
+
+    background: $primary-color;
+
+    animation: blobs 15s ease-in-out infinite alternate;
```
grafit-off, line 85: You can use placeholders or mixins for repeated properties like animation here.

src/styles/blocks/_watch.scss
```diff
@@ -0,0 +1,87 @@
...
+  margin-top: 50px;
+
+  &__circle {
+    @include circle(600px);
+
+    background: linear-gradient(-45deg, #ee7752, #e73c7e, #23a6d5, #23d5ab);
+    background-size: 400% 400%;
+
+    animation: blobs 15s ease-in-out infinite alternate,
+      bg-gradient 15s ease infinite;
```
grafit-off, line 15: You can use variables for animation duration, for the reason that this value is the same

src/styles/blocks/_watch.scss
```diff
@@ -0,0 +1,87 @@
...
+    transition: 2s;
```
grafit-off, line 16: You should set transition properties that are going to change, so the browser doesn't need to watch all properties

src/styles/utils/_keyframes.scss
```diff
@@ -0,0 +1,77 @@
...
+  50% {
+    transform: rotate(180deg);
+  }
+  75% {
+    transform: rotate(270deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
grafit-off, line 53: In my opinion, you can make a mixin for this case

grafit-off (approved):
Good job 🔥 , awesome animation 👍

src/styles/blocks/_watch.scss
Denchepornyuk: yep, you can't add margin to BEM block, remove it

src/styles/blocks/_watch.scss
Denchepornyuk: agree, create variable sounds good

src/styles/blocks/_watch.scss
Denchepornyuk: agree, let's fix it

Denchepornyuk (changes requested):

Anna-Tkachenko (approved):
Awesome! ⭐️

PR #1487, opened 2022-08-23
timer

Ivanenko1402:
- [DEMO LINK](https://Ivanenko1402.github.io/layout_stop-watch/)


alexandra-protyanova (changes requested):
Something wrong with your demo, haven't u forget to run `npm run deploy`?

src/styles/main.scss
```diff
@@ -0,0 +1,9 @@
+* {
+  margin: 0;
```
liliia-livandovskaya, line 2: its forbidden to use `margin: 0;` in `*`

PR #1499, opened 2022-08-24
Layout was done

OleksandrGlulhovskiy:
- [DEMO LINK](https://OleksandrGlulhovskiy.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,17 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="watch">
+      <div class="watch__hands"></div>
+      <div class="watch__center"></div>
+    </div>
+  </body>
 </html>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,14 +1,17 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="watch">
+      <div class="watch__hands"></div>
```
imaksweb, line 13: You can make arrows by using ::before and ::after of watch div

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,28 @@
...
+
+@mixin positioning {
+  @include circle(25px);
+
+  position: absolute;
+  top: 50%;
+  left: 50%;
+
+  transform: translate(-50%, -50%);
+}
```
imaksweb, line 15: It doesn't look like reusable stuff. As for me it'll be better to create a placeholder with centering functionality

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,45 @@
...
+      animation-duration: 1s;
+    }
+  }
+
+  &__center {
+    @include positioning;
+
+    z-index: 1;
+
+    border-radius: 50%;
```
imaksweb, line 32: You already have this rule inside your positioning mixin

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,28 @@
...
+  position: absolute;
+  top: 5px;
+  left: 15px;
+
+  border-radius: 5px;
+
+  transform-origin: left;
+
+  animation: circle infinite linear;
+}
```
imaksweb, line 28: I suppose you can use placeholder instead of mixin

imaksweb (changes requested):
Not bad! Some things look a bit complicated. Also I've noticed a strange behavior, if you open a devTools and hover over watch div you see something strange with your div width. And one more thing it's not good that your watch div has 0 height.

src/index.html
OleksandrGlulhovskiy: I think it's not a mistake, but you are right

OleksandrGlulhovskiy (commented):
What about watch height, I think all is right, when I change size off my brouser window my watch does not change

MaksimusCRS (approved):
GJ

PR #1504, opened 2022-08-24
Created animated stop-watch

Dikey945:
- [DEMO LINK](https://Dikey945.github.io/layout_stop-watch/)


src/styles/blocks/page.scss
```diff
@@ -0,0 +1,12 @@
+.page {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+
+  &__watch {
+    position: relative;
+    height: $watch-size;
+    width: $watch-size;
```
ArturNagaycev, line 10: i think we have to move width and height to block stop-watch. because when we move block to another place - our watch will be without these properties

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,10 @@
+$circle-size: 40px;
+$circle-colore: #000;
+$millisecond-color: #f00;
+$second-arrow-color: #00f;
+$watch-size: 570px;
+$milisecond-arrow-duration: 1s;
+$second-arrow-duration: 60s;
+$arrow-width: 20px;
+$millisecond-lengh: 250px;
+$second-length: $millisecond-lengh / 1.5;
```
ArturNagaycev, line 10: Orphographic mistakes. I see that you write the same word in different ways (like "color and color", "length and length", "milisecond and millisecond"). Your future colleagues can have some problems with it.

ArturNagaycev (changes requested):
i liked your code and watch )) but left a few little comments ))

src/styles/utils/_variables.scss
Dikey945: Agree) sometimes orphographic my bad side)

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,32 @@
...
+  background: url("../images/clock_scale.png") center / cover no-repeat;
+
+  height: $watch-size;
+  width: $watch-size;
+
+  &__center {
+    @extend %position;
+    @include circle($circle-size);
+    z-index: 1;
+    border-radius: 50%;
```
OleksiiBlezniuk, line 11: You've already set it in mixin 🙂

OleksiiBlezniuk (approved):
Nice UI!

PR #1505, opened 2022-08-24
watch was created

vallerii:


src/styles/block/watch.scss
```diff
@@ -0,0 +1,66 @@
...
+    min-height: 250px;
+    width: 20px;
+    border-radius: 50%;
+    margin: auto;
+
+    transform-origin: bottom;
+    animation: move-minute;
+    animation-iteration-count: infinite;
+    animation-duration: 60s;
+    animation-timing-function: linear;
```
artem-musii, line 32: You can combine all this properties into one

src/styles/block/watch.scss
```diff
@@ -0,0 +1,66 @@
...
+}
+
+@keyframes move-seconds {
+  0% {
+    transform: rotate(180deg);
+  }
+  100% {
+    transform: rotate(540deg);
+  }
+}
```
artem-musii, line 66: You can use animation one time for two arrows

src/styles/block/watch.scss
vallerii: sure,thanks

src/styles/block/watch.scss
artem-musii: yes, but if you used position: relative for watch__content and position: absolute for your arrows, you would be able to use only one animation, but maybe it's not so important

artem-musii (approved):

Anna-Tkachenko (approved):
Well Done! ⭐️

PR #1510, opened 2022-08-25
created stop-watch using animation

Apollonia5:
- [DEMO LINK](https://Apollonia5.github.io/layout_stop-watch/)


src/styles/block/stop-watch.scss
```diff
@@ -0,0 +1,49 @@
...
+    animation-duration: 1s;
+    @extend %animation-char;
+  }
+}
+
+@keyframes spin {
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
AndriyZhabynets, line 49: I would recommend using a separate file for the entire animation

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,18 @@
...
+
+$container-size: 300px;
+$point-size: 20px;
+
+$point-margin: ($container-size - $point-size) / 2;
+
+$minutes-width: 16px;
+$minutes-height: 100px;
+$minutes-margin-top: $container-size/ 2 - $minutes-height;
+$minutes-margin-left: ($container-size - $minutes-width) / 2;
```
AndriyZhabynets, line 13: I would recommend using a function to calculate these variables

AndriyZhabynets (approved):
I have some recommendations for you, but I think they are optional, so i approve this task

src/styles/utils/_variables.scss
Apollonia5: These are two different formulas. I think it is not necessary to create complicated forms.

PR #1511, opened 2022-08-25
Stop watch

glugovsky:
- [DEMO LINK](https://glugovsky.github.io/layout_stop-watch/)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,43 @@
+.clock {
+  @extend %flex-center;
+  @include circle(640px);
+
+  background-image: url("../images/clock.svg");
+  box-shadow: 4px 4px 16px #ccc;
+
+  transition-property: transform, box-shadow;
+  transition-duration: 0.3s;
+  transition-timing-function: ease-out;
```
artem-musii, line 10: Just a tip) You can combine all transition properties into one

artem-musii (approved):
GJ, everything looks fine for me

src/styles/blocks/clock.scss
glugovsky: Yes, but than I would need to repeat the duration and the timing function for each property and I am not sure that would be less repetitive, considering that both transitions share these values

MaksimusCRS (approved):
GJ

PR #1512, opened 2022-08-25
Initial commit

xiimiik:
- [DEMO LINK](https://xiimiik.github.io/layout_stop-watch/)


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,34 @@
+.main {
```
illia-sosnytskyi, line 1: each bem block should be in separate file

MaksimusCRS (approved):
GJ

PR #1528, opened 2022-08-26
clock was created

varkovroma:
[DEMO LINK](https://varkovroma.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,70 @@
+.clock {
+  position: absolute;
+  height: $clock-size;
+  width: $clock-size;
+  background: #fff;
+  border-radius: 100%;
+  border: 10px solid #535a5a;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  margin: auto;
+
+  &__center {
+    height: 6px;
+    width: 6px;
+    background-color: #262626;
+    position: absolute;
+    border-radius: 100%;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    margin: auto;
+  }
+
+  &__minutes {
+    width: 4px;
+    height: 100%;
+    position: absolute;
+    left: 0;
+    right: 0;
+    margin: 0 auto;
+    animation: time 60s infinite linear;
+
+    &::before {
+      position: absolute;
+      content: "";
+      background-color: #262626;
+      height: 60px;
+      width: 4px;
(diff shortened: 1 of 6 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,66 @@
...
+      top: 50px;
+    }
+  }
+}
+
+@keyframes move {
+  to {
+    transform: rotate(360deg);
+  }
+}
```
vovchuk-nazar, line 66: Good job

vovchuk-nazar (approved):

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,66 @@
+.clock {
+  @include circle(200px);
+  position: absolute;
+  background: #fff;
+  border: 10px solid #535a5a;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  margin: auto;
```
OleksiiBlezniuk, line 10: Don't use margins or position on BEM blocks

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,66 @@
...
+
+  &__center {
+    @include circle(6px);
+    background-color: #262626;
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
```
OleksiiBlezniuk, line 19: I think it can be moved to placeholder

src/styles/blocks/clock.scss
OleksiiBlezniuk: Move animations to separate file

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,12 @@
+* {
+  margin: 0;
+  padding: 0;
```
OleksiiBlezniuk, line 3: Don't use * for default styles reset

OleksiiBlezniuk (changes requested):

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,56 @@
+.clock {
+  @include circle(200px);
+  position: absolute;
```
OleksiiBlezniuk, line 3: Do not use positioning props for BEM block (you can use only `position: relative` if some of child elements have `position: absolute` in case if it's used only for that purpose)

OleksiiBlezniuk (approved):

PR #1529, opened 2022-08-26
Task was done like clock with minutes and seconds arrows

SanyaKovalenko:
[DEMO LINK](https://SanyaKovalenko.github.io/layout_stop-watch/)

### DIFF
--- a/src/ulils/_mixins.scss
+++ b/src/ulils/_mixins.scss
@@ -1,0 +1,37 @@
+@mixin watch_position {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+}
+
+@mixin seconds-arrow_gabarits {
+  width: $seconds-arrow_width;
+  height: $seconds-arrow_height;
+}
+
+@mixin minutes-arrow_gabarits {
+  width: $minute-arrow_width;
+  height: $minute-arrow_height;
+}
+
+@mixin point_gabarites {
+  width: $point_size;
+  height: $point_size;
+  border-radius: 50%;
+}
+@mixin arrows_position {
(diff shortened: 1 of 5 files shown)


src/ulils/_mixins.scss
```diff
@@ -0,0 +1,37 @@
...
+
+@mixin point_gabarites {
+  width: $point_size;
+  height: $point_size;
+  border-radius: 50%;
+}
+@mixin arrows_position {
+  position: absolute;
+  top: 0;
+  left: 10px;
```
silchyn, line 25: What is 10

src/ulils/_mixins.scss
```diff
@@ -0,0 +1,37 @@
...
+}
+
+@mixin point_position {
+  position: absolute;
+  top: 0;
+  left: 0;
+}
+
+@mixin default-arrows_position {
+  transform-origin: 0 10px;
```
silchyn, line 35: What is 10

silchyn (changes requested):
I am pretty sure those are not second and minute hands, and what means that delay at the beginning for one arrow?

SanyaKovalenko:
> I am pretty sure those are not second and minute hands, and what means that delay at the beginning for one arrow?

It was my idea to let arrows move like clock and delay was added because only after one ritate of the seconds arrow the minutes one should move but maybe it was uncorrect.

silchyn (changes requested):
Understood, but is there such delay in real clock?

Anna-Tkachenko (approved):

PR #1550, opened 2022-08-29
Animated stopwatch created

roman-zherebetskyi:
- [DEMO LINK](https://roman-zherebetskyi.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,7 +10,7 @@
   </head>
   <body class="page">
     <div class="watch">
-      <div class="center-point"></div>
+      <div class="watch__center-point"></div>
     </div>
   </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,16 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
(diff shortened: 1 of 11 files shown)


src/index.html
```diff
@@ -1,14 +1,16 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body class="page">
+    <div class="watch">
+      <div class="center-point"></div>
```
VikaMartin22, line 13: You should use BEM naming, like: watch > watch__center-point

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,4 @@
+@mixin watch-size($page-height) {
```
VikaMartin22, line 1: It would be better to use abstract name for this mixin
'Cause you use it not only for watch container.

VikaMartin22 (commented):
Good work, small naming corrections

VikaMartin22 (approved):

src/styles/utils/_mixins.scss
roman-zherebetskyi: I'm agree with you

MaksimusCRS (approved):
GJ

PR #1553, opened 2022-08-30
stopwatch

Arsen-Shpak:
[DEMO LINK](https://Arsen-Shpak.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,25 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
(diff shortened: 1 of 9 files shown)


src/index.html
```diff
@@ -1,14 +1,25 @@
...
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body class="page">
+    <!-- <div class="watch">
+      <div class="watch__container"></div>
+      <div class="watch__second"></div>
+      <div class="watch__minute"></div>
+      <div class="watch__center"></div>
+    </div> -->
```
roman-zherebetskyi, line 17: hahahah why do you need this comment? but acctually I'd prefer this one than that below

src/styles/blocks.scss/watch.scss
```diff
@@ -0,0 +1,62 @@
+.container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  position: relative;
```
roman-zherebetskyi, line 5: your watch.scss is file with all styles for BEM block you can't use positioning for it. So your <div class='container'> is your BEM block because __border, __center e.t.c. are it's elements. If you want to use positioning then wrap your watch into some other block.

or you can use this .container like your wrap container, but then don't bound your watch block with it

src/styles/blocks.scss/watch.scss
```diff
@@ -0,0 +1,62 @@
...
+      width: $big-hand-width;
+
+      background-color: $minute-color;
+      border-radius: 10px;
+
+      animation: rotate 60s linear infinite;
+      transform-origin: bottom;
+    }
+  }
+  &__second {
```
roman-zherebetskyi, line 34: it would be good to make a blank line here

roman-zherebetskyi (changes requested):
Good job! but still, try to fix some BEM bugs ;)

Anna-Tkachenko (approved):

PR #1555, opened 2022-08-30
layout_stop-watch is running

Denys-Grynyshyn:
- [DEMO LINK](https://Denys-Grynyshyn.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/stopwatch/stopwatch.scss
+++ b/src/styles/stopwatch/stopwatch.scss
@@ -1,0 +1,41 @@
+.main {
+  @include horizontal-center-flex;
+
+  min-height: 100vh;
+
+  margin: 0;
+}
+
+.stopwatch {
+  @include horizontal-center-flex;
+  @include circle($watch-size);
+
+  position: relative;
+  border: 15px solid $border-watch-color;
+  background-color: $watch-background-color;
+
+  &:hover {
+    cursor: pointer;
+  }
+
+  &:hover > * {
+    animation-play-state: paused;
+  }
+
+  &__center {
+    @include circle(20px);
+
+    position: absolute;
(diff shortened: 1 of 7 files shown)


src/styles/stopwatch/stopwatch.scss
```diff
@@ -0,0 +1,41 @@
+.main {
+  @include horizontal-center-flex;
+
+  min-height: 100vh;
+
+  margin: 0;
+}
```
roman-zherebetskyi, line 7: better to move this styles to individual .scss file to make your BEM block more independent

src/styles/stopwatch/stopwatch.scss
```diff
@@ -0,0 +1,41 @@
...
+
+.stopwatch {
+  @include horizontal-center-flex;
```
roman-zherebetskyi, line 10: your 'horizontal-center-flex' contains the margin but it's not allowed to use margins for BEM blocks

src/styles/stopwatch/stopwatch.scss
```diff
@@ -0,0 +1,41 @@
...
+  @include circle($watch-size);
+
+  position: relative;
```
roman-zherebetskyi, line 13: you can't use positioning for BEM blocks

roman-zherebetskyi (commented):
Nice job! but try to fix those couple small things

src/styles/stopwatch/stopwatch.scss
grafit-off: @roman-zherebetskyi In this case, he can for a reason that this does not affect the neighbouring blocks, it's just to position children's elements

PR #1574, opened 2022-09-05
Stopwatch was created

IrynaK-a:
- [DEMO LINK](https://IrynaK-a.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/stopwatch.scss
+++ b/src/styles/blocks/stopwatch.scss
@@ -1,0 +1,44 @@
+.stopwatch {
+  position: relative;
+  margin-bottom: 20px;
+  border: 3px solid black;
+  @include stopwatch-size();
+
+  &__center-point {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+
+    width: 8%;
+    height: 8%;
+    border-radius: 50%;
+    border: 2px solid red;
+  }
+
+  &__minute-hand {
+    @extend %hands-border;
+    @include hands-position;
+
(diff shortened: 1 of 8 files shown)


Bohdan-Maksymiuk (changes requested):
> Create a working stopwatch with **minute** and **second** hands using only CSS animations.

so choose right animation duration

IrynaK-a:
> > Create a working stopwatch with **minute** and **second** hands using only CSS animations.
> 
> so choose right animation duration

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,44 @@
...
+  &__second-hand {
+    @extend %hands-border;
+    @include hands-position;
+
+    margin: auto;
+    height: 49%;
+    width: 0;
+    border-color: black;
+
+    animation: stopwatch 1s linear 1s infinite;
```
ArturErmolenko, line 41:
```suggestion
    animation: stopwatch 60s linear 1s infinite;
```
1s is the time of a full rotation of the hand. To simulate a stop-watch  here should be seconds

ArturErmolenko (changes requested):

src/styles/blocks/stopwatch.scss
IrynaK-a: > 1s is the time of a full rotation of the hand. To simulate a stop-watch here should be seconds

Sorry, but I still confused.
I have two hands: one with 1s duration other with 60s. They should be 60s each?

PR #1625, opened 2022-10-03
Develop

Mykola-Busarhin:
- [DEMO LINK](https://Mykola-Busarhin.github.io/layout_stop-watch/)

Is it OK that when I hover over one of the hands, it stops, but the other is still moving? Or is it a bug? Can`t figure out how to hover over the two hands at the same time.


src/styles/main.scss
```diff
@@ -0,0 +1,123 @@
+$watch-width: 400px;
```
alexandra-protyanova, line 1: move to separate file variables & keyframe

alexandra-protyanova (approved):
Good job!

PR #1628, opened 2022-10-04
Watch was created

iryna-lelyk:
- [DEMO LINK](https://iryna-lelyk.github.io/layout_stop-watch/)


src/styles/utilits/_mixin.scss
```diff
@@ -0,0 +1,9 @@
+@mixin centerDotSize ($size) {
+  width: $size;
+  height: $size;
+}
+
+@mixin positionOfHand ($hand-height) {
+  top: calc(50% - #{$hand-height});
+  left: calc(50% - (#{$hand-width} / 2));
+}
```
Illia-Skladnik, line 9: But you don't need to use calc in scss

Illia-Skladnik (approved):

src/styles/utilits/_mixin.scss
iryna-lelyk: Sometimes yes, but here i have different units of measurement (px and %), so without calc i cannot make a calculation.

PR #1650, opened 2022-10-07
Init

RomeoKs:


src/index.html
```diff
@@ -1,14 +1,37 @@
...
+      </div>
+
+      <div class="stopwatch__minutes"></div>
+
+      <div class="stopwatch__seconds"></div>
+
+    </div>
+  </body>
+
+</html>
```
mgrinko, line 37:
```suggestion
  <body>
    <div class="stopwatch">
      <div class="stopwatch__hands">
        <div class="stopwatch__hands-minutes"></div>
        <div class="stopwatch__hands-seconds"></div>
        <div class="stopwatch__hands-circle"></div>
      </div>

      <div class="stopwatch__minutes"></div>
      <div class="stopwatch__seconds"></div>
    </div>
  </body>
</html>

```

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  @include flex-justify-align(center, center);
+}
+
+.stopwatch {
+  width: 500px;
+  height: 500px;
+  background-color: aqua;
+  border-radius: 50%;
+  @include flex-justify-align(center, center);
```
mgrinko, line 13: I don't think such a mixin simplifies the code
I would just copy the rules

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,65 @@
...
+
+    &-circle {
+      width: 20px;
+      height: 20px;
+      background-color: yellow;
+      border-radius: 50%;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-8px, -10px);
```
mgrinko, line 28: Where did you get this exact pixels?
I think it should be -50%

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,65 @@
...
+  background-color: aqua;
+  border-radius: 50%;
+  @include flex-justify-align(center, center);
+
+  &__hands {
+    width: 400px;
+    height: 400px;
+    position: relative;
+
+    &-circle {
```
mgrinko, line 20: don't do this!
circle is not a part of __hands but independent element
It will be hard to find it by full name __hands-circle

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,65 @@
...
+      position: absolute;
+      left: 50%;
+      top: 30px;
+      width: 5px;
+      height: 170px;
+      box-sizing: border-box;
+      background-color: rgb(0, 174, 255);
+      animation: rotate steps(60) var(--hour) infinite;
+      transform-origin: center bottom;
+      z-index: 2;
```
mgrinko, line 42: If we put elements in correct order we don't need to set exact z-index

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,5 @@
+:root {
+  --second: 1s;
+  --minute: calc(var(--second) * 60);
+  --hour: calc(var(--minute) * 60);
+}
```
mgrinko, line 5: sass variable are simpler if you are not going to change them using javascript

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    &-seconds {
+      position: absolute;
+      left: calc(50% + 1.5px);
+      width: 2px;
```
mgrinko, line 48: Don't use such a magic pixels

mgrinko (changes requested):

src/styles/blocks/stopwatch.scss
RomeoKs: -10px is the actual 50% and -8px the 50% minus the half of hands-minutes (except the 0.5px - magic pixel)

Without the 2px it looks like this - https://imgur.com/a/srcmoeY

src/styles/utils/_variables.scss
RomeoKs: Here I'll definitely need some explaining

src/index.html
```diff
@@ -1,14 +1,34 @@
...
+    <div class="stopwatch">
+      <div class="stopwatch__hands">
+        <div class="stopwatch__hands-minutes"></div>
+        <div class="stopwatch__hands-seconds"></div>
+
+      </div>
+      <div class="stopwatch__minutes"></div>
+      <div class="stopwatch__seconds"></div>
+      <div class="stopwatch__circle"></div>
+    </div>
```
mgrinko, line 31:
```suggestion
    <div class="stopwatch">
      <div class="stopwatch__hands">
        <div class="stopwatch__hands-minutes"></div>
        <div class="stopwatch__hands-seconds"></div>
      </div>
      
      <div class="stopwatch__minutes"></div>
      <div class="stopwatch__seconds"></div>
      <div class="stopwatch__circle"></div>
    </div>
```

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,69 @@
...
+  justify-content: center;
+  align-items: center;
+
+
+  &__hands {
+    width: 400px;
+    height: 400px;
+    position: relative;
+
+    &-minutes {
```
mgrinko, line 26: put &__hands-minutes separately
because otherwise, it is not possible to search by the class name

you should nest only element that are connected by sense but not just with the name prefix

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,69 @@
...
+      position: absolute;
+      left: 50%;
+      top: 30px;
+      width: 5px;
+      height: 170px;
```
mgrinko, line 31: I don't understand where do these numbers come from
Are they somehow related to the size of the watch itself (500px)

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,69 @@
...
+
+  &__circle {
+    width: 20px;
+    height: 20px;
+    background-color: yellow;
+    border-radius: 50%;
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-40%, -50%);
```
mgrinko, line 67: I don't understand why you use 40% but not 50%
It is not clear how this number is calculated

src/styles/utils/_variables.scss
mgrinko: calc works in browser, but sass is calculated beforehand
so if you don't need this calculation to relate on some `vm` or `%` please use SCSS variables

mgrinko (changes requested):

src/styles/utils/_variables.scss
RomeoKs: Thank you for explaining

src/styles/blocks/stopwatch.scss
RomeoKs: Yep, my logic failed.
I should have checked in the Dev Tools which elements are properly center before doing any calculations.

In my mind the circle was off and the hands were centered, but it was vice versa.

I really appreciate you pointing me where I'm failing.

Thank you!

src/styles/blocks/stopwatch.scss
RomeoKs: Seconds hands are smaller in height with 30px than the Minutes hands.

But, I've changed them to make more sense.

mgrinko (approved):
I like it now!

PR #1654, opened 2022-10-07
stop-watch is done

Svitlana-Yudina:
- [DEMO LINK](https://Svitlana-Yudina.github.io/layout_stop-watch/)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,56 @@
+@keyframes move {
+  from {
+    transform: rotate(0);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
```
karyna-salnykova, line 8: i think it would be better to make another file (in utils) for keyframes

src/styles/main.scss
```diff
@@ -0,0 +1,15 @@
...
+@import "./blocks/clock.scss";
+
+.clock-position {
+  display: flex;
+  height: 100vh;
+
+  &__clock {
+    margin: auto;
+  }
+}
```
karyna-salnykova, line 15: i think it should be in clock.scss

src/styles/utils/reset.scss
```diff
@@ -0,0 +1,10 @@
+body,
+p,
+h1,
+h2,
+section,
+ul,
+a {
+  margin: 0;
+  padding: 0;
+}
```
karyna-salnykova, line 10: maybe you should use * to reset all elements on the page?

mykola-koval (approved):

PR #1703, opened 2022-10-27
fixed lint bugs

MaksymBaturinskyi:
- [DEMO LINK](https://gabrielmaksimovich.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,77 @@
...
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    animation: sec 3600s steps(60, end) infinite;
+
+    &::after {
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -100%);
```
vadiimvooo, line 44: It would be better if you were use here `@mixin` instead of dublicating styles. (The same styles you have in lines 60-63)

vadiimvooo (changes requested):
According to requirements you must change your file structure. You haven't to write all styles in main.scss, so create separate files to each BEM block.

src/styles/main.scss
GabrielMaksimovich: thanks, i will change it ,still on a catalog hovers project)

vadiimvooo (approved):
Good job!)

src/styles/main.scss
GabrielMaksimovich: done

solaryasha (approved):
Nicely done!

PR #1721, opened 2022-11-03
Done

SalehAkaJim:
- [DEMO LINK](https://SalehBUD.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,117 @@
+.clock {
+  @include defaultCenter;
+  width: $clockSize;
+  height: $clockSize;
+  border-radius: $borderRadius;
+  border: 2px solid $mainColor;
+
+  &__number-3 {
+    position: absolute;
+    width: ($clockSize - 60);
+    color: $itemsColor;
+    transform: rotate(180deg);
+    @include selectChild(180deg);
+  }
+
+  &__number-6 {
+    position: absolute;
+    width: ($clockSize - 60);
+    color: $itemsColor;
+    transform: rotate(270deg);
+    @include selectChild(90deg);
+  }
+
+  &__number-9 {
+    position: absolute;
+    width: ($clockSize - 40);
+    color: $itemsColor;
+    transform: rotate(360deg);
+    @include selectChild(10deg);
+  }
+
+  &__number-12 {
+    position: absolute;
+    width: ($clockSize - 60);
+    color: $itemsColor;
+    transform: rotate(450deg);
+    @include selectChild(-90deg);
+  }
+
+  &__second {
+    @include defaultCenter;
+    position: absolute;
+    padding: $secondPadding;
+    border-radius: $borderRadius;
+    animation: timeSecondAnimation 60s infinite linear;
+    z-index: 2;
+
+    &-line {
+      position: absolute;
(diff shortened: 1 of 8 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,117 @@
...
+}
+
+@keyframes timeHourAnimation {
+  from {
+    transform: rotate(135deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
```
mgrinko, line 117: There should be the same animation but with different durations

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,10 @@
+@import url("https://fonts.googleapis.com/css2?family=Poor+Story&display=swap");
+
+$mainColor: #3E3D4E;
+$clockSize: 360px;
+$itemsColor: #FCFCFC;
+$borderRadius: 50%;
```
mgrinko, line 6: I don't think this should go to a variable if you use it once

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,10 @@
...
+
+$secondPadding: 3.8px;
+$minutePadding: 5.7px;
+$hourPadding: 7.6px;
```
mgrinko, line 10: hm...
I don't like decimal pixels

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,117 @@
...
+
+  &__minute {
+    @include defaultCenter;
+    position: absolute;
+    padding: $minutePadding;
+    border-radius: $borderRadius;
+    animation: timeMinuteAnimation 1440s infinite linear;
+    z-index: 1;
+
+    &-line {
```
mgrinko, line 65: It is not a good idea to use scss nesting just not to write the same prefix again
The main purpose of nesting is to nest some related things like modifiers or pseudo-classes but not just the text prefix just because it is the same now

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,117 @@
...
+    padding: $secondPadding;
+    border-radius: $borderRadius;
+    animation: timeSecondAnimation 60s infinite linear;
+    z-index: 2;
+
+    &-line {
+      position: absolute;
+      left: ($secondPadding * 2);
+      height: 1px;
+      width: ($clockSize / 2.6);
```
mgrinko, line 52: why 2.6 exactly?

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,13 @@
+@mixin defaultCenter {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
```
mgrinko, line 5: I would never guess that `defaultCenter` mixin adds display flex
I don't think this mixin is really helpful
I'd better duplicate the code

mgrinko (changes requested):

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,7 @@
+@mixin selectChild($deg) {
+  div {
```
mgrinko, line 2: I don't think div is a good selector in a mixin

mgrinko (approved):
Good job!

PR #1727, opened 2022-11-05
Watch

llllewvvaa:
- [DEMO LINK](https://llllewvvaa.github.io/layout_stop-watch/)


src/index.html
```diff
@@ -1,14 +1,18 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body class="page">
+    <div class="watch watch--screen">
+      <div class="watch__minutes-hand"></div>
+      <div class="watch__seconds-hand"></div>
+      <div class="watch__dot"></div>
+    </div>
```
mr-valentos, line 16: p1 Don't forget about semantics, div development is not a better option 
`<section>`
remove `<div class="watch__dot"></div>` if you don't use it

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,12 @@
...
+$color-second-hand: #00539cff;
+$left-hand-position: 250px;
+$width-minute-hand: 170px;
+$height-minute-hand: 9px;
+$width-second-hand: 220px;
+$height-second-hand: 6px;
+$duration-minute-hand: 3600s;
+$duration-second-hand: 60s;
+$watch-height: 500px;
+$watch-width: 500px;
```
mr-valentos, line 12: p.10 But don't overuse them, don't create variable for the value that's used just once. (optional)

mr-valentos (changes requested):
st review

src/index.html
llllewvvaa: **Fixed**

PR #1740, opened 2022-11-08
task solved

Lun4trick:
- [DEMO LINK](https://lun4trick.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,24 @@
+.clock {
+  margin: 25vh auto;
+
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: $clockSize;
+  height: $clockSize;
+
+  border-radius: $circle;
+  background-color: #5b6264;
+  animation: shape 5s linear 0s infinite alternate;
+
+  &__middle {
+    position: absolute;
+    display: flex;
+    width: 50px;
+    height: 50px;
+
+    border-radius: $circle;
+    background-color: #2c2a2a;
+  }
+
+}
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,4 @@
+@import "./utils/variables";
+@import "./utils/animations";
+@import "./utils/loops";
+@import "./blocks/clock.scss";
--- a/src/styles/utils/_loops.scss
+++ b/src/styles/utils/_loops.scss
@@ -1,0 +1,28 @@
+@each $hand, $animTime in $hourHands {
+
+  .clock__#{$hand} {
+
+    position: absolute;
+    margin-bottom: $handPosition;
+    height: $handPosition;
+    width: 10px;
+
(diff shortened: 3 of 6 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,24 @@
+.clock {
+  margin: 25vh auto;
```
nurdanokcu, line 2: As you can see in the [link](https://snipboard.io/0Ny4jH.jpg) the clock isn't in the center and there is a scroll in the page. You can solve this problem by giving display flex to parent and setting height to viewport height (for the parent.)

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,24 @@
...
+  width: $clockSize;
+  height: $clockSize;
+
+  border-radius: $circle;
+  background-color: #5b6264;
+  animation: shape 5s linear 0s infinite alternate;
+
+  &__middle {
+    position: absolute;
+    display: flex;
```
nurdanokcu, line 16: These 2 properties aren't needed. You already centered it with parent element flex properties.  Code works fine without them.

src/styles/main.scss
```diff
@@ -0,0 +1,4 @@
+@import "./utils/variables";
+@import "./utils/animations";
+@import "./utils/loops";
+@import "./blocks/clock.scss";
```
nurdanokcu, line 4:
```suggestion
@import "./utils/variables";
@import "./utils/animations";
@import "./utils/loops";

@import "./blocks/clock.scss";
```

Blank line would be nice here to separate utils from blocks.

src/styles/utils/_loops.scss
```diff
@@ -0,0 +1,28 @@
+@each $hand, $animTime in $hourHands {
+
+  .clock__#{$hand} {
+
+    position: absolute;
```
nurdanokcu, line 5:
```suggestion
@each $hand, $animTime in $hourHands {
  .clock__#{$hand} {
    position: absolute;
```

No need for a blank line here.  Just a  suggestion though. 😊

src/styles/utils/_loops.scss
```diff
@@ -0,0 +1,28 @@
...
+    transform-origin: bottom;
+  }
+
+}
+
+@each $number, $marginDirection in $numbers {
+
+  .clock__number-#{$number} {
+    
+    margin-#{$marginDirection}: 400px;
```
nurdanokcu, line 22: Same blank line suggestions.

nurdanokcu (changes requested):
Amazing implementation with loops, animations and variables. 👏
I only asked you to center the clock in the middle of the page. And I added same spacing suggestions.

nurdanokcu (approved):
Suggestions are considered.  Very nice work.

naziamoff (approved):
Well done 🔥 🚀

PR #1753, opened 2022-11-10
added stop-watch

romanblyashuk94:
- [DEMO LINK](https://romanblyashuk94.github.io/layout_stop-watch/)


src/styles/blocks/stop-watch.scss
```diff
@@ -0,0 +1,66 @@
...
+    top: 50%;
+    left: 50%;
+    z-index: 4;
+    transform: translate(-50%, -50%);
+
+    width: 5%;
+    height: 5%;
+
+    border-radius: 50%;
+    background-color: black;
```
Nazar-Brunarskyi, line 22: u use 50% quite a lot, so maybe it would be a good idea to create a variable (change it everywhere)

```suggestion
    top: variable;
    left: variable;
    z-index: 4;
    transform: translate(-variable, -variable);

    width: 5%;
    height: 5%;

    border-radius: 50%;
    background-color: black;
```

Nazar-Brunarskyi (approved):

src/styles/blocks/stop-watch.scss
romanblyashuk94: I don't think so.
This values are independent.
Position calculates from parrent positional block
Transform/translate calculates from self size

I am not sure but let's leave it up to the mentor

thank you for review)

src/styles/blocks/stop-watch.scss
```diff
@@ -0,0 +1,66 @@
...
+
+  }
+
+  &__hand {
+    position: absolute;
+    left: 50%;
+    bottom: 50%;
```
ivan-kadykalo, line 29: Create a mixin or placeholder for repetitive parameters.

ivan-kadykalo (changes requested):

ivan-kadykalo:
Good job and cool design🚀

solaryasha (approved):
🔥

PR #1757, opened 2022-11-10
first commit

Bir3791:
- [DEMO LINK](https://Bir3791.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/_variables.scss
+++ b/src/styles/_variables.scss
@@ -1,0 +1,29 @@
+$side : 200px;
+$duration-hour : 216000s;
+$duration-minute : 3600s;
+$duration-second : 60s;
+$duration-microsecond : 1s;
+$background-primary : #eee;
+$clock-hand-primary: #000;
+
+%hands {
+  position: absolute;
+  height: 4px;
+  background-color: black;
+  border-radius: 5px;
+  top: 50%;
+  left: 50%;
+  transform-origin: left;
+}
+
+%position {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+}
+
+%clock-hand {
+  background-color: #{$clock-hand-primary};
+  border-radius: 10px;
+}
--- a/src/styles/easy-clock.scss
+++ b/src/styles/easy-clock.scss
@@ -1,0 +1,40 @@
+.easy-clock {
+  position: relative;
+  box-sizing: border-box;
+  width: #{$side};
+  height: #{$side};
+  background-color: #{$background-primary};
+  border: 0.5px solid black;
+  border-radius: 50%;
+
+  &__hour {
+    @extend %hands;
+    width: 50px;
+    transform: rotate(1deg);
+    animation: move-clock #{$duration-hour} linear infinite;
+  }
+  &__minute {
+    @extend %hands;
+    width: 80px;
+    animation: move-clock #{$duration-minute} linear infinite;
+  }
+  &__second {
+    @extend %hands;
+    width: 40px;
+    animation: move-clock #{$duration-second} linear infinite;
+  }
+  &__microsecond {
+    @extend %hands;
+    width: 20px;
+    animation: move-clock #{$duration-microsecond} linear infinite;
+  }
+}
+
+@keyframes move-clock {
+  from {
+    transform: rotate(-90deg);
+  }
+  to {
+    transform: rotate(269deg);
+  }
+}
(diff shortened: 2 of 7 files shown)


liliia-livandovskaya (changes requested):
Deploy the project Screenshot 2022-11-10 at 17 20 55

src/styles/_variables.scss
```diff
@@ -0,0 +1,29 @@
...
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+}
+
+%clock-hand {
+  background-color: #{$clock-hand-primary};
+  border-radius: 10px;
+}
```
SanyaBratashchuk, line 29: move it to the seperate file

src/styles/easy-clock.scss
```diff
@@ -0,0 +1,40 @@
+.easy-clock {
+  position: relative;
+  box-sizing: border-box;
+  width: #{$side};
+  height: #{$side};
```
SanyaBratashchuk, line 5:
```suggestion
  width: $side;
  height: $side;
```
Fix it everywhere

src/styles/easy-clock.scss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+@keyframes move-clock {
+  from {
+    transform: rotate(-90deg);
+  }
+  to {
+    transform: rotate(269deg);
+  }
+}
```
SanyaBratashchuk, line 40: move keyframes to the seperate file 
```suggestion
@keyframes hand-rotation {
  from {
    transform: rotate(-90deg);
  }
  to {
    transform: rotate(269deg);
  }
}
```
Why not from 0 to 360?

src/styles/watch.scss
```diff
@@ -0,0 +1,83 @@
...
+}
+
+@keyframes move {
+  from {
+    transform: rotate(-90deg);
+  }
+  to {
+    transform: rotate(270deg);
+  }
+}
```
SanyaBratashchuk, line 83: Try to use only one keyframe

SanyaBratashchuk (changes requested):

src/styles/_variables.scss
Bir3791: added placeholder.scss

src/styles/easy-clock.scss
Bir3791: fixed

src/styles/easy-clock.scss
Bir3791: because it is usually horizontal block and hand on 0deg will be 3o'clock. animation should start from 12 o'clock so -90deg. -90 + 359 = 269deg

SanyaBratashchuk: try `transofrm-origin: bottom`

src/styles/watch.scss
```diff
@@ -0,0 +1,74 @@
...
+    left: 98.5px;
+    animation: move-clock $duration-minute linear infinite;
+  }
+
+  &__second {
+    @extend %clock-hand;
+    position: absolute;
+    width: 75px;
+    height: 2px;
+    transform-origin: 1px 1px;
```
SanyaBratashchuk, line 58:
```suggestion
    transform-origin: 1px 1px;
```
check it maybe try different value

src/styles/watch.scss
```diff
@@ -0,0 +1,74 @@
...
+    left: 98.5px;
+    animation: move-clock $duration-second linear infinite;
+  }
+
+  &__microsecond {
+    @extend %clock-hand;
+    position: absolute;
+    width: 32px;
+    height: 1px;
+    transform-origin: 0.5px 0.5px;
```
SanyaBratashchuk, line 69: the same

SanyaBratashchuk (changes requested):
Try transform-origin bottom

Bir3791:
I made two watches. in the first one I tried to make the rotation point perfect to the surrounding elements.  it is not possible to do this, because different hands has differents height. and that's why in "watch" i have size like 1px or 0.5px. 

realising this is not correct, I added another clock where there is no such problem, but the hands of clock do not move perfectly.

so "Watch"it's about perfectly geometric without easy change size.
"easy clock" about small cod, good and easy change size without perfect geometry and different watch-hands.

src/styles/watch.scss
Bir3791: i need point of rotation the same like in other clock hands

SanyaBratashchuk (approved):

PR #1762, opened 2022-11-10
watch

Sveta-Kryukova:
- [DEMO LINK](https://Sveta-Kryukova.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/style.scss
+++ b/src/styles/blocks/style.scss
@@ -9,9 +9,9 @@
   &__small-hand {
     position: absolute;
     left: $left-position;
-    top: 190px;
-    height: 4px;
-    width: 140px;
+    top: $distance-to-center-for-small-hand;
+    height: $width-of-hand;
+    width: $small-hand-size;
     border-radius: 3px;
     background-color: rgba(255, 255, 255, 0.6);
     animation: turn 3600s linear infinite ;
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -1,14 +1,17 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body class="container">
+    <div class="watch">
+      <div class="watch__small-hand"></div>
+      <div class="watch__big-hand"></div>
+    </div>
+  </body>
```
4ebupel, line 16: The only change I can suggest is that:  Checklist p.1 - Don't forget about semantics, div development is not a better option; I don't think it is really matters in such small project but checklist is a checklist, so it's better to change that
<sub>Anyway I'll approve that 'cause I don't consider this as a really big mistake or something</sub>

4ebupel (approved):
Wunderbar!)

src/styles/blocks/style.scss
```diff
@@ -0,0 +1,38 @@
...
+  position: relative;
+  height: $watch-size;
+  width: $watch-size;
+  background-image: url("../image/watch.png");
+  background-position: center;
+  border-radius: 50%;
+
+  &__small-hand {
+    position: absolute;
+    left: $left-position;
```
ivan-kadykalo, line 11: why not 50%?

src/styles/blocks/style.scss
```diff
@@ -0,0 +1,38 @@
...
+    top: 190px;
```
ivan-kadykalo, line 12: What will be if we will change clock size?

ivan-kadykalo (changes requested):

src/styles/blocks/style.scss
Sveta-Kryukova: get it, thanks

solaryasha (approved):
👍

ivan-kadykalo (approved):

PR #1763, opened 2022-11-10
Stopwatch task done

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/stopWatch.scss
+++ b/src/styles/blocks/stopWatch.scss
@@ -1,0 +1,76 @@
+.stopWatch {
+  display: block;
+
+  &__frame {
+    width: $clock-size;
+    height: $clock-size;
+
+    border: 5px double $color-brownish;
+    box-shadow: 5;
+    border-radius: 100%;
+
+    position: relative;
+    margin: 130px auto;
+
+    background-color: whitesmoke;
+  }
+
+  &__stick {
+    position: absolute;
+    left: 0;
+    right: 0;
+    bottom: $clock-size / 2;
+    margin: 0 auto;
+    transform-origin: bottom center;
+
+    &--minute {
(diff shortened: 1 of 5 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,21 @@
+@import "/styles/utils/variables.scss";
+@import "/styles/blocks/stopWatch.scss";
+
+body {
+  background: linear-gradient(-45deg, #ee7752, #e73c7e, #23a6d5, #23d5ab);
+  background-size: 200% 200%;
```
SalehBUD, line 6: Just a fix on background 😊
```suggestion
  background-size: 300% 300%;
  background-repeat: no-repeat;
```

SalehBUD: Btw, you can move all of this general codes (the ones you're applying on html, body, ...) into:
`utils -> _extends.scss`
and then import it in here...

SalehBUD (changes requested):

SalehBUD (approved):
😊

mmahfuzi:
> 😊

Thanks Saleh :)

src/styles/blocks/stopWatch.scss
```diff
@@ -0,0 +1,71 @@
...
+    position: absolute;
+    left: 0;
+    right: 0;
+    bottom: $clock-size / 2;
+    margin: 0 auto;
+    transform-origin: bottom center;
+
+    &--minute {
+      @include objSize(8px, 135px);
+      z-index: 9;
```
naziamoff, line 26: why z-index is 9, but not let's say 1 here?

naziamoff (approved):
Good job! Just one question

PR #1767, opened 2022-11-11
created stop-watch with animations

rmpolonskiy:
https://rmpolonskiy.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/blocks/stopwatch.scss
+++ b/src/styles/blocks/stopwatch.scss
@@ -1,0 +1,74 @@
+.stopwatch {
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  @include circle($stopwatch-size);
+  margin: 0 auto;
+  border: 5px solid $stopwatch-color;
+  cursor: pointer;
+
+  &:hover .animated {
+    animation-play-state: paused;
+  }
+
+  &__seconds {
+    z-index: 1;
+    width: 250px;
+    height: $arrow-height;
+    margin-left: -50px;
+
+    background-color: $stopwatch-color;
+    border-radius: $arrow-border-radius;
+
+    transform-origin: center right;
+    transform: rotate(90deg);
+    animation: seconds
+      60s
+      $arrow-timing-f
+      60
+      $arrow-direction
+      $arrow-fill-mode;
+  }
+
+  &__minutes {
+    width: 200px;
+    height: $arrow-height;
+    background-color: darkblue;
+    border-radius: $arrow-border-radius;
+
+    transform-origin: center left;
+    transform: rotate(-90deg);
+    animation: minutes
+      3600s
+      $arrow-timing-f
+      1
+      $arrow-direction
+      $arrow-fill-mode;
+  }
+
+  &__dot {
+    position: absolute;
+    z-index: 1;
+    @include circle($dot-size);
+    background-color: $stopwatch-color;
+  }
+}
+
+@keyframes seconds {
+  from {
+    transform: rotate(90deg);
+  }
+  to {
+    transform: rotate(450deg);
+  }
+}
+
+@keyframes minutes {
+  from {
+    transform: rotate(-90deg);
+  }
+  to {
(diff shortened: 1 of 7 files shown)


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,74 @@
...
+}
+
+@keyframes minutes {
+  from {
+    transform: rotate(-90deg);
+  }
+  to {
+    transform: rotate(270deg);
+  }
+}
```
SanyaBratashchuk, line 74: Try to unify this animations

SanyaBratashchuk: And move it to the seperate file

SanyaBratashchuk (changes requested):

src/styles/utils/_animations.scss
```diff
@@ -0,0 +1,11 @@
...
+  to {
+    transform: rotate(450deg);
+  }
+}
+
+@keyframes minutes {
+  to {
+    transform: rotate(270deg);
+  }
+}
```
SanyaBratashchuk, line 11:
```suggestion
@keyframes rotate {
  from {
    transform: rotate(0)
  }
  
  to {
    transform: rotate(360deg)
  }
}
```
Make it work use transform-origin with correct value

SanyaBratashchuk (changes requested):

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,59 @@
+.stopwatch {
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  @include circle($stopwatch-size);
+  margin: 0 auto;
+  border: 5px solid $stopwatch-color;
+  cursor: pointer;
```
ivanchuk-yurii, line 9: do you need this?

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,59 @@
...
+
+  &:hover .animated {
```
ivanchuk-yurii, line 11: strange logic)

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,59 @@
...
+      3600s
+      $arrow-timing-f
+      1
+      $arrow-direction
+      $arrow-fill-mode;
+  }
+
+  &__dot {
+    position: absolute;
+    top: 47.5%;
```
ivanchuk-yurii, line 53: don't use magic numbers, calculate position basing on dot size

ivanchuk-yurii (changes requested):
sorry for the extra comments)

src/styles/blocks/stopwatch.scss
rmpolonskiy: Why not?) 
There was mentioned "There are no strict design you should follow for this task, so you can choose on your own."

src/styles/blocks/stopwatch.scss
rmpolonskiy: Why?

src/styles/blocks/stopwatch.scss
rmpolonskiy: Ok, I will try to change it today

src/styles/blocks/stopwatch.scss
rmpolonskiy: As it is stopwatch, the logic is that we should stop it at any moment. 
So, I decided to realise that in this way (when we hover on this stopwatch, the animation stoped). That's why I implemented cursor:pointer, because user shoud see, that this element is interactive

rmpolonskiy: What's wrong with this logic? 
Explain your comment, plz

Bohdan-Maksymiuk: better to have some button to toggle start, it's not obvious)

Bohdan-Maksymiuk (approved):

PR #1769, opened 2022-11-11
Develop

Rijo-J:
- [DEMO LINK](https://rijo-j.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,145 @@
+.clock {
+  position: relative;
+
+  @extend %flexedCentered;
+
+  background-color: $colorClockBackground;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,145 @@
...
+      content: "";
+
+      @include squared(10px);
+
+      border-radius: 50%;
+      background-color: $colorMarkers;
+    }
+
+    &:nth-of-type(3n + 3) {
+      width: 92%;
```
naziamoff, line 45: looks like a magic number here
Why exactly 92%, but not 93%?

naziamoff (approved):
GJ!

PR #1773, opened 2022-11-11
Develop

sercansimsek:


IoghiStudio (changes requested):
You have to fix your tests.

sercansimsek:
> 
To be honest do not know how to fix it :D but I did get same check fail in my previous tasks too I do not know why exactly but Roman said that it is okay to each of them.

meforoma (approved):

PR #1786, opened 2022-11-12
watch

Lav1004:
- [DEMO LINK](https://Lav1004.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/style.scss
+++ b/src/styles/blocks/style.scss
@@ -1,0 +1,67 @@
+body {
+  margin: 0;
+  padding: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-height: 100vh;
+  background: #000;
+}
+
+.container {
+  width: calc(8rem + 15px);
+  height: calc(8rem + 15px);
+  border-radius: 50%;
+  background: linear-gradient(-230deg, rgb(100, 6, 252), green, blue);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.clock {
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 8rem;
+  height: 8rem;
+  background: #000;
+  border-radius: 50%;
+}
+
+span {
+  width: 0.7rem;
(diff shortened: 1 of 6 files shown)


src/styles/blocks/style.scss
```diff
@@ -0,0 +1,67 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 8rem;
+  height: 8rem;
+  background: #000;
+  border-radius: 50%;
+}
+
+span {
```
TarasHoliuk, line 32: it's unclear what is it. set some class like `clock__center` and use class selector in CSS instead of `span`

src/styles/blocks/style.scss
```diff
@@ -0,0 +1,67 @@
...
+  width: 0.7rem;
+  height: 0.7rem;
```
TarasHoliuk, line 34: why everything in `rem`?

src/styles/blocks/style.scss
```diff
@@ -0,0 +1,67 @@
...
+}
+
+.clock::before, .clock::after {
+  position: absolute;
+  content: '';
+  top: calc(4rem * 0.25);
+  width: 0.4rem;
+  background: linear-gradient(green, blue);
+  border-radius: 8px;
+  transform-origin: center calc(100% - calc(0.4rem / 2));
```
TarasHoliuk, line 47: Interesting decision to use `::before` and `::after` but it would be easier to set it if you create separate elements for arrows.

src/styles/blocks/style.scss
```diff
@@ -0,0 +1,67 @@
...
+  animation: anim infinite linear;
+}
+
+.clock::before {
+  height: 3.2rem;
+  animation-duration: 10s; 
+}
+
+.clock::after {
+  top: calc(4rem * 0.25 + 1.6rem);
```
TarasHoliuk, line 57: there are a lot of "magic" numbers, try to create variables for some of them

TarasHoliuk (changes requested):

PR #1791, opened 2022-11-13
completed task

Alireza-Mohammadhossein:


IoghiStudio (approved):
Awesome!

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,7 @@
+@mixin animation {
+  animation-direction: nomal;
+  animation-iteration-count: infinite;
+  animation-name: clock;
+}
+
+$w-h : 500px;
```
naziamoff, line 7: better to avoid 1-letter variables by all costs.
Let's come up with sth like $clock-size or something

naziamoff (approved):
Great job! 🚀

PR #1850, opened 2022-12-02
solved the task

MeryamWazeen:
- [DEMO LINK](https://meryamwazeen.github.io/layout_stop-watch/)


src/styles/blocks/animation.css
```diff
@@ -0,0 +1,10 @@
+@-webkit-keyframes time {
+  to {
+    transform: rotate(350deg);
+  }
+}
+@keyframes time {
+  to {
+    transform: rotate(350deg);
+  }
+}
```
polosanya, line 10: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

polosanya (changes requested):
Please don't ignore the linter and fix all the problems.
image
Also, clock hands do not start from the centre of the clock. It looks a bit strange, try to fix it too
image

src/styles/blocks/_stopwatch.scss
```diff
@@ -0,0 +1,55 @@
...
+  left: 0;
+  right: 0;
+  border: 10px solid rgb(127, 255, 212);
+  background-color: #fff;
+  border-radius: 50%;
+  margin: auto;
+  box-shadow: 20px 20px 10px 15px rgba(0, 0, 255, 0.2);
+}
+
+.stopwatch__center {
```
jstmpelowycz, line 16: - `center` itself should be placed in front of all the other elements (hint – use `z-index` property)
- let's use nesting

src/styles/blocks/_stopwatch.scss
```diff
@@ -0,0 +1,55 @@
...
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: 0;
```
jstmpelowycz, line 20: move it up by the half of `center` height (i.e. radius, radius = height / 2) – you should place right in the middle of `stopwatch` itself (hint – use variables)

src/styles/blocks/_stopwatch.scss
```diff
@@ -0,0 +1,55 @@
...
+  position: absolute;
+  left: 0;
+  right: 0;
+  margin: auto;
+  top: 50px;
+  width: 15px;
+  height: 200px;
+  background-color: #4c9aaa;
+  border-radius: 60px;
+  animation: move 60s steps(60, end) infinite;
```
jstmpelowycz, line 39: when configuring `animation` property in such cases where movement is linear (no speed deviations), don't forget indicating it explicitly
```suggestion
animation: move 60s steps(60, end) infinite linear;
```

src/styles/blocks/_stopwatch.scss
```diff
@@ -0,0 +1,55 @@
...
+  transform-origin: bottom;
+}
+
+.stopwatch__minute {
```
jstmpelowycz, line 43: make sure to fix the styles of this element the same way as I've mentioned in the previous comments

jstmpelowycz (changes requested):
great job! let's apply a few last hints:
- fix the issues left down below
- when you create a file dedicated to some BEM-block, you call it `<element_name>.scss`, when it's a file dedicated to placing something like variables, mixins, placeholders, resetting/normalizing styles, you add double underscores `__`

one more suggestion – take care of how you organize your styles – [this](- https://9elements.com/css-rule-order/) article is handy for you

src/styles/blocks/_stopwatch.scss
MeryamWazeen: done

src/styles/blocks/_stopwatch.scss
MeryamWazeen: i have used z-index so now the .center is covering the watch hands.

src/styles/blocks/_stopwatch.scss
MeryamWazeen: i didn't understand your this comment ! :(

src/styles/blocks/_stopwatch.scss
```diff
@@ -0,0 +1,56 @@
+.stopwatch {
```
polosanya, line 1: You can call this file just `stopwatch.scss`. No need to underline if it's a file for BEM block

src/styles/blocks/_stopwatch.scss
polosanya: I don't see that it's fixed
image

src/styles/main.scss
```diff
@@ -0,0 +1,5 @@
+
```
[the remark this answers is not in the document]
polosanya: Not fixed

src/styles/blocks/_stopwatch.scss
```diff
@@ -0,0 +1,56 @@
...
+  position: absolute;
+  left: 0;
+  right: 0;
+  margin: auto;
+  top: 100px;
+  width: 10px;
+  height: 150px;
+  background-color: #d15659;
+  border-radius: 60px;
+  animation: move 3600s infinite;
```
polosanya, line 54: Would be great to create variables for colors and animation time

src/styles/utls/_variabels.scss
```diff
@@ -0,0 +1 @@
+$mainColor: rgb(0,128,128);
```
polosanya, line 1: Only one variables? I think you can create more for most important values

polosanya (changes requested):
Hi, I see you haven't fixed previous comments, maybe you did not push your changes?
If you have any questions, please ask them in the Slack chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review 🙂

qoa11a (approved):

PR #1853, opened 2022-12-03
added watch

Malva37:
- [DEMO LINK](https://Malva37.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/minute.scss
+++ b/src/styles/blocks/minute.scss
@@ -1,0 +1,8 @@
+.minute {
+  background-color: chartreuse;
+  width: $hans-width;
+  height: ($watch-size / 2) * 0.8;
+  transform-origin: center 0;
+  animation: move steps(60) $hour infinite;
+
+}
--- a/src/styles/utils/_variables.scss
+++ b/src/styles/utils/_variables.scss
@@ -1,0 +1,15 @@
+$watch-size:500px;
+$hans-width:4px;
+$second: 1s;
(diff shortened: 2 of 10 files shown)


src/styles/blocks/minute.scss
```diff
@@ -0,0 +1,8 @@
+.minute {
+  background-color: chartreuse;
+  width: $hans-width;
+  height: ($watch-size / 2) * 0.8;
+  transform-origin: center 0;
+  animation: move steps(60) $hour infinite;
+
```
Anna-Tkachenko, line 7: delete redundant line

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
+$watch-size:500px;
+$hans-width:4px;
```
Anna-Tkachenko, line 2:
```suggestion
$watch-size: 500px;
$hans-width: 4px;
```

src/styles/blocks/minute.scss
```diff
@@ -0,0 +1,8 @@
+.minute {
+  background-color: chartreuse;
+  width: $hans-width;
+  height: ($watch-size / 2) * 0.8;
```
Anna-Tkachenko, line 4: it's not clear what is 0.8 and 0.95 for the second class, will be better to create variables for these values

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1856, opened 2022-12-04
stop-watch v1.0

Donovanoff:
[DEMO LINK](https://donovanoff.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,18 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -1,14 +1,18 @@
...
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="clock">
+      <div class="clock__minutes"></div>
+      <div class="clock__hour"></div>
```
vlad-demchuk, line 14: if it's `minutes` and `hour` so let's set corresponding time of animations

src/styles/main.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+
+@mixin square ($size) {
+  width: $size;
+  height: $size;
+}
+
+body {
+  margin: 0;
+}
```
vlad-demchuk, line 21: move it to separate files e.g `variables, mixins, reset and keyframes`

vlad-demchuk (changes requested):

src/index.html
Donovanoff: I'm start with minutes and hours but then I understood that its too long to check it

src/styles/main.scss
Donovanoff: okay

vlad-demchuk (approved):

PR #1862, opened 2022-12-06
watch was done

yuriiKorolov:
- [DEMO LINK](https://yuriiKorolov.github.io/layout_stop-watch/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,319 +4,333 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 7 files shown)


qoa11a (changes requested):
Minute and second hands should move respectively. Minute hand should make one full rotate every 60 minutes. Second hand should make one full rotate every 60 seconds.

AlisaVlasova (changes requested):
hands should start from the same point
image

PR #1871, opened 2022-12-08
add stopwatch

NVyrsta:
- [DEMO LINK](https://NVyrsta.github.io/layout_stop-watch/)


src/main.js
```diff
@@ -1 +1,16 @@
...
+btn.addEventListener('click', (event) => {
+  if (event.target.innerText === 'PAUSE') {
+    event.target.innerText = 'START';
+  } else {
+    event.target.innerText = 'PAUSE';
+  }
+
+  secondHand.classList.toggle('play-animation');
+  msHand.classList.toggle('play-animation');
+});
```
OlhaYevstifieieva, line 16: This solution is ingenious!

OlhaYevstifieieva (approved):
Perfect job! ✨ I love it 🙌🏻

src/main.js
NVyrsta: Thanks:)

Denchepornyuk (approved):
Screenshot 2022-12-13 at 16 34 42

a great initiative with button 🔥
Probably text should be changed, when the time is running should be `pause`

PR #1872, opened 2022-12-09
Make watch

sv-dubov:
- [DEMO LINK](https://sv-dubov.github.io/layout_stop-watch/)


src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,47 @@
...
+    transform-origin: bottom center;
+  }
+
+  &__center {
+    position: absolute;
+    top: 46.25%;
+    height: 7.5%;
+    width: 7.5%;
+    background-color: $watch-color-main;
+    border-radius: 50%;
```
Sergyurch, line 40: It would be good to make a mixin for this. You repeat the same properties in each selector.

src/index.html
```diff
@@ -1,14 +1,18 @@
...
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="watch watch--position">
```
Sergyurch, line 12: Watch--position is not good name for modifier. Modifier should describe a special state of the block or the element. It's better to use parent element to position a block. Something like page__watch.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,47 @@
...
+  }
+}
+
+.watch--position {
+  position: relative;
+  margin: auto;
+}
```
Sergyurch, line 47: Modifier can't set external space or position for an element. Use parent element.

Sergyurch (changes requested):
Change modifier to parent element.

src/styles/blocks/watch.scss
sv-dubov: I've done it but now it looks like harder to read code.

src/index.html
sv-dubov: I've just made it in the "watch" class.

sv-dubov: Done

ArturErmolenko (approved):

PR #1880, opened 2022-12-11
Simple clock was added

svitjojo:
- [DEMO LINK](https://svitjojo.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/stopWatch.scss
+++ b/src/styles/blocks/stopWatch.scss
@@ -1,0 +1,58 @@
+.stopWatch {
+  position: relative;
+  height: 500px;
+  width: 500px;
+  background-color: #fff;
+  border: 3px solid #000;
+  border-radius: 50%;
+  display: flex;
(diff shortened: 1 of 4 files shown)


oleh-holovnykh (changes requested):
Well done! Looks clear) I propose to slightly reduce the amount of code

src/styles/blocks/stopWatch.scss
```diff
@@ -0,0 +1,45 @@
...
+    background-color: #000;
+    z-index: 2;
+  }
+
+  &__second {
+    transform-origin: bottom;
+    animation: rotation 60s linear infinite;
+    position: absolute;
+    bottom: 50%;
+    left: 247.5px;
```
Denchepornyuk, line 20: float numbers like `247.5` are not good practice, try to always use integers

Denchepornyuk (approved):

PR #1891, opened 2022-12-13
Performed the task

maasiimka:


src/styles/components/clock.scss
```diff
@@ -0,0 +1,49 @@
...
+
+    transform-origin: bottom center;
+
+    width: 14px;
+    height: 300px;
+
+    background-color: rgb(115, 115, 255);
+    border-radius: 7px;
+
+    animation: rotate 60s linear infinite;
```
svitjojo, line 30: you put wrong time, it moving too fast
```suggestion
    animation: rotate 3600s linear infinite;
```

src/styles/components/clock.scss
```diff
@@ -0,0 +1,49 @@
...
+
+    transform-origin: bottom center;
+
+    width: 14px;
+    height: 400px;
+
+    background-color: rgb(255, 163, 179);
+    border-radius: 7px;
+
+    animation: rotate 1s linear infinite;
```
svitjojo, line 47: same here, too faast
```suggestion
    animation: rotate 60s linear infinite;
```

svitjojo (changes requested):
It's so funny how fast your clock rotating 😁

src/styles/components/clock.scss
maasiimka: Ah, really, it was a bit stupid. Thank you for review

src/styles/components/clock.scss
```diff
@@ -0,0 +1,49 @@
...
+  &__center {
+    position: relative;
+
+    width: 30px;
+    height: 30px;
+
+    border: 2px solid rgb(115, 115, 255);
+    border-radius: 50%;
+    background-color: #fff;
+    box-shadow: 0 0 0 2px #fff;
```
Denchepornyuk, line 13: if you use hash for colors `#fff` in the code you should use it in all places, don't combine rgb and hash in the styles

src/styles/components/clock.scss
```diff
@@ -0,0 +1,49 @@
...
+
+    transform-origin: bottom center;
+
+    width: 14px;
+    height: 300px;
+
+    background-color: rgb(115, 115, 255);
+    border-radius: 7px;
+
+    animation: rotate 3600s linear infinite;
```
Denchepornyuk, line 30: would be great to move some numbers in variables

Denchepornyuk (changes requested):
Screenshot 2022-12-15 at 18 36 04

clock should in the middle of the page

src/styles/components/clock.scss
maasiimka: Maybe I should better make placeholder am mixin for arrows?  I already pushed it so you can look at this

src/styles/utils/_placeholders.scss
```diff
@@ -0,0 +1,10 @@
+%arrows-properties {
+  position: absolute;
+  left: 8px;
+  bottom: 15px;
```
Denchepornyuk, line 4: good solution with mixin and placeholder

Denchepornyuk (approved):

PR #1902, opened 2022-12-14
watch first commit

Bogdan-Kutasevich:
- [DEMO LINK](https://Bogdan-Kutasevich.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,74 @@
+.watch {
+  position: relative;
+  height: $watch-size;
+  width: $watch-size;
+  background-color: $watch-background-color;
+  border-radius: 50%;
+  border: $watch-border-size solid $watch-background-color;
+
+  &__numbers {
+    position: absolute;
+    font-size: $number-fs;
+    transform: translate(-50%, -50%);
+    color: $color-defailt-white;
+    &--number {
+      &-12 {
+        left: 50%;
+      }
+      &-3 {
+        left: 100%;
+        top: 50%;
+      }
+      &-6 {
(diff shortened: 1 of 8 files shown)


src/styles/blocks/wrapper.scss
```diff
@@ -0,0 +1,5 @@
+.wrapper {
+  display: flex;
+  justify-content: center;
+  padding: 100px;
```
GarryRocksman, line 4: its hardcode vertical positioning ) use height: 100vh; and align-items: center;

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,74 @@
+.watch {
```
GarryRocksman, line 1: You can add to hands one general class like - "watch_ hand" and put all of the repeated code there.  Here a lot of repeated code.  Hand positioning is almost the same.

GarryRocksman (changes requested):
You did great work. But.  Check styles for repeated code/ you can optimize it if you add one general class for "hands"

src/styles/blocks/watch.scss
Bogdan-Kutasevich: There i have diffrent properties for hands, like width and height, so positions for theese element are diffrent and related with it size. Only transform-origin the same for this element.

src/styles/blocks/wrapper.scss
Bogdan-Kutasevich: Make sense, thanks ;)

alexandra-protyanova (approved):
Good job 👍

PR #1909, opened 2022-12-14
Stop watch was created

andrii-andriushchenko:
- [DEMO LINK](https://andrii-andriushchenko.github.io/layout_stop-watch/)

Have a good mood! :)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,7 +8,7 @@
     <title>Stop watch</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
+  <body class="content">
     <div class="watches">
       <div class="watches__center"></div>
       <div class="watches__second-arrow"></div>
--- a/src/styles/blocks/content.scss
+++ b/src/styles/blocks/content.scss
@@ -1,0 +1,6 @@
+.content {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,5 +1,5 @@
 @import "./utils/variables";
 @import "./utils/mixins";
 
-@import "./normalize.scss";
+@import "./blocks/content.scss";
 @import "./blocks/watches.scss";


src/styles/blocks/watches.scss
```diff
@@ -0,0 +1,46 @@
...
+    height: $watch-middle-height;
+    top: 50%;
+    left: 50%;
+    border-radius: 50%;
+    background-color: $item-bgc;
+    transform: translate(-50%, -50%);
+  }
+
+  &__minute-arrow {
+    @include similarItemValues;
```
GarryRocksman, line 23: good that you group similar values in @mixin, but you are using it twice. In this case, it is useless. Just add one generic class like "watch__hand" and put these styles there

GarryRocksman: ` animation: rotate 60s linear infinite;` also you can put there

src/styles/normalize.scss
```diff
@@ -0,0 +1,6 @@
+body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
```
GarryRocksman, line 6: just add the class "page" to the body and move your styles here. Normalize file is generally used for fonts etc.

GarryRocksman (changes requested):
You did great work, just go through my comments and do some fixes in the code. But it's just my opinion so take it easy )))

andrii-andriushchenko:
> You did great work, just go through my comments and do some fixes in the code. But it's just my opinion so take it easy )))

Thanks a lot 🚀

src/styles/normalize.scss
andrii-andriushchenko: Thanks. Fixed

src/styles/blocks/watches.scss
andrii-andriushchenko: As I understand acc BEM it is necessary to create new block, but I dont know is it correct, because I cannot to reuse block with this characteristics. I'll ask about it, thanks

alexandra-protyanova: is there still an issue? now it looks correct according to BEM

alexandra-protyanova (approved):
Well done 🔥

PR #1914, opened 2022-12-14
watch task

BrandonSunrise:
- [DEMO LINK](https://BrandonSunrise.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/utils/_variable.scss
+++ b/src/styles/utils/_variable.scss
@@ -1,0 +1,14 @@
+$watch-size: 500px;
+$center-size: $watch-size / 20;
+$color-main: black;
+$color-second: red;
+
+@keyframes rotate {
+  from {
+    transform: rotate(0);
+  }
+
+  to {
+    transform: rotate(1turn);
+  }
+}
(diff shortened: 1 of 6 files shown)


src/styles/utils/_variable.scss
```diff
@@ -0,0 +1,14 @@
...
+
+@keyframes rotate {
+  from {
+    transform: rotate(0);
+  }
+
+  to {
+    transform: rotate(1turn);
+  }
+}
```
GarryRocksman, line 14: i don't think that its right place for animation

GarryRocksman (approved):
GJ, just check the right place for animation

src/styles/utils/_variable.scss
Denchepornyuk: agree, isn't the best place, you can keep it in dial.scss

Denchepornyuk (changes requested):
Screenshot 2022-12-15 at 18 42 37

hide arrows behind the circle

src/styles/utils/_variable.scss
BrandonSunrise: done!

Anna-Tkachenko (approved):
Good job!

PR #1917, opened 2022-12-14
Stop-watch_first commit

olena-kapustiuk:


src/styles/reset.scss
```diff
@@ -0,0 +1,4 @@
+* {
+  margin: 0;
+  padding: 0;
```
vlad-demchuk, line 3: it's better not to reset properties for all elements (`*`)

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,98 @@
+.clock {
+  display: flex;
+  position: relative;
```
vlad-demchuk, line 3: checklist
image

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,98 @@
...
+
+  @keyframes sweep {
+    from {
+      transform: rotate(0deg);
+    }
+
+    to {
+      transform: rotate(360deg);
+    }
+  }
```
vlad-demchuk, line 52: move to a separate file

vlad-demchuk (changes requested):
Looks good, just fix comments

AlisaVlasova (approved):
awesome

PR #1919, opened 2022-12-14
Develop

ihor-soloviov:
- [DEMO LINK](https://ihor-soloviov.github.io/layout_stop-watch/)


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,62 @@
...
+    width: $stopwatch-size;
+    margin: auto;
+    display: flex;
+  }
+
+  &__center {
+    height: $item-size;
+    width: $item-size;
+    background-color: #555;
+    border-radius: 50%;
```
anna-inozemets, line 20: You can create a mixin for making circle

src/styles/utils/_functions.scss
```diff
@@ -0,0 +1,19 @@
+@keyframes moveSec {
+  0% {
+    transform: #{$second-translate} rotate(0);
+  }
+
+  100% {
+    transform: #{$second-translate} rotate(360deg);
+  }
+}
```
anna-inozemets, line 9: I think that it is possible just use rotate and transform-origin property. I am not sure, but I think you can create just one animation with propertu "rotate: and use different duration in different hands

anna-inozemets (approved):
Well  done)

src/styles/blocks/stopwatch.scss
Denchepornyuk: agree, mixin looks good here

src/styles/utils/_functions.scss
Denchepornyuk: yeah, it's possible to make this animation only with rotate

denys-cheporniuk (changes requested):
Screenshot 2022-12-16 at 09 25 37

arrows should be hidden behind a circle

Anna-Tkachenko (approved):
Good job! ⭐️

PR #1921, opened 2022-12-14
added layout_stop-watch

Dashaverz:
https://Dashaverz.github.io/layout_stop-watch/


OlyaZaiets:
According to the task, the stopwatch display should be in the center of the page.

Denchepornyuk (approved):

PR #1938, opened 2022-12-18
Fixed some mistakes

SashaViun:
- [DEMO LINK](https://SashaViun.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,42 @@
+.watch {
+  @include clock;
+  background-color: black;
+  width: $watch-size;
+  height: $watch-size;
+  margin-top: 50px;
+  border: 4px solid $main-color;
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  cursor: pointer;
+
+  &__center {
+    @include design;
+    @include clock;
+    height: $center-size;
+    width: $center-size;
+  }
+
+  &__seconds {
+    @include design;
+    width: 1%;
+    height: 36%;
+    top: 14%;
+    transform-origin: bottom;
+    animation-name: seconds;
+    animation-duration: 60s;
+    animation-iteration-count: 60;
+  }
+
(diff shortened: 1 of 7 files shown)


PolinaSukhorukova:
Good job! 🔥

PolinaSukhorukova (changes requested):
Second hand unnaturally slows down at the end of the minute.  Try to add animation-timing-function.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,42 @@
...
+    @include clock;
+    height: $center-size;
+    width: $center-size;
+  }
+
+  &__seconds {
+    @include design;
+    width: 1%;
+    height: 36%;
+    top: 14%;
```
Anna-Tkachenko, line 25: it's not clear what these value means
will be better to calculate them somehow depending on the size of watch / hands

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,42 @@
...
+    animation-name: seconds;
+    animation-duration: 60s;
+    animation-iteration-count: 60;
+  }
+
+  &__minutes {
+    @include design;
+    height: 30%;
+    width: 2%;
+    top: 20%;
```
Anna-Tkachenko, line 36: same here

Anna-Tkachenko (changes requested):
Why did you set border-radius 45% for the main container? let's make watch the circle, not some other figure

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
+@keyframes minutes {
+  100% {
+    transform: rotate(360deg);
+  }
+}
+
+@keyframes seconds {
```
alexandra-protyanova, line 7: You have 2 absolutely the same keyframes, you can use one of them and nothing will change

alexandra-protyanova (approved):
Good job 👍

PR #1943, opened 2022-12-21
Stopwatch task

IvanLugovskiy:
- [DEMO LINK](https://IvanLugovskiy.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,99 @@
...
+  margin: 0;
+}
+
+.clock {
+  display: block;
+  position: relative;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: $margin-value;
+  margin-bottom: $margin-value;
```
Anastasiia-Svintsova, line 14:
```suggestion
  margin: $margin-value auto
```

Anastasiia-Svintsova, line 14: The checklist 
image

src/styles/main.scss
```diff
@@ -0,0 +1,99 @@
...
+    transform: translate(-50%, -50%);
+    border-radius: 50%;
+    z-index: 1;
+  }
+
+  &__minute-arrow,
+  &__secund-arrow {
+    position: absolute;
+    top: $position-margin;
+    left: $position-margin;
```
Anastasiia-Svintsova, line 47: Maybe you should create a mixin for it. You use it more than once

src/styles/variables.scss
```diff
@@ -0,0 +1,12 @@
+$clock-width: 500px;
+$clock-height: 500px;
+$main-height: 100%;
```
Anastasiia-Svintsova, line 3: Seems that `$main-height` is a useless variable

src/styles/variables.scss
```diff
@@ -0,0 +1,12 @@
+$clock-width: 500px;
+$clock-height: 500px;
```
Anastasiia-Svintsova, line 2:
```suggestion
$clock-size: 500px;
```

Anastasiia-Svintsova (changes requested):

src/styles/main.scss
IvanLugovskiy: Probably I am not understanding something, could you explain to me please?

src/styles/variables.scss
IvanLugovskiy: why useless ? I used it for positioning in the center in "calc"

src/styles/main.scss
loralevitska: You have bad and good examples 
image
do you read [checklist](https://github.com/mate-academy/layout_stop-watch/blob/master/checklist.md)?

You should create extra class for this div and another for main to show that your div is an element and not a block. If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review

loralevitska (changes requested):
almost done)

fix last comment
image

src/styles/main.scss
IvanLugovskiy: Oh, sorry, sure! I read check list, and it's clear to me in example, but I was not sure where I should apply this rule in my case( ( I also understood about fe_chat

qoa11a (approved):

PR #1948, opened 2022-12-23
clock was created

KirillMaslov:
[DEMO LINK](https://KirillMaslov.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,52 @@
...
+    height: 20px;
+    width: 20px;
+    background-color: aqua;
+    border-radius: 50%;
+    z-index: 2;
+  }
+
+  &__minute:hover {
+    transform: rotate(180deg);
+  }
```
ViktorHavryliuk, line 47: Clock must work all the time, not only on hover. Use css animations and keyframes instead of hover and transition

src/styles/main.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  &__second {
+    background-color: red;
+    height: $clock__second-size;
+    position: absolute;
+    width: $clock__arrows-height;
+    border-radius: 10px;
+    transition: transform 1s linear;
+    transform-origin: bottom center;
+  }
```
ViktorHavryliuk, line 35: Move rules that are common for both arrows into mixin and than include it here, to avoid code duplication

ViktorHavryliuk (changes requested):

ViktorHavryliuk (approved):

PR #1953, opened 2022-12-26
stop watch

BeeGooDee:
https://BeeGooDee.github.io/layout_stop-watch/


DarkMistyRoom (changes requested):
It is a creative solution, but we can't see the visible animation. I understand that you use animation to implement it, but you change the content, no visual animation. you can add some flipping effect, or implement the version with hands

ViktorHavryliuk (approved):

PR #1958, opened 2022-12-28
first commit

toljan1:
- [DEMO LINK](https://toljan1.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,59 @@
...
+.second {
+  width: 15px;
+  position: absolute;
+  bottom: 50%;
+  left: 50%;
+  height: 360px;
+  border-radius: 10px;
+  background-color: #f00;
+  transition: transform;
+  transform-origin: bottom;
```
witflash, line 46: You can use BEM and create the block and elements, and add some modifiers to avoid rules duplicate:
```
<div class="clock">
  <div class="clock__arrow clock__arrow--minutes></div>
  <div class="clock__circle></div>
  <div class="clock__arrow clock__arrow--seconds></div>
</div>
```

witflash (changes requested):
Hey!
Check the comments, please!

PR #1966, opened 2023-01-02
basic clock added, indicator offset glitch, why indicators are unevenly placed and mismatch the second indicator?

domkab:
- [DEMO LINK](https://domkab.github.io/layout_stop-watch/)

I still didnt put the code properly into sass syntax, bem blocks, but will do that after you help me fix the issue with clock indicators :)))

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,92 @@
 <!doctype html>
-<html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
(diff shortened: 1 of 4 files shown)


qoa11a (changes requested):
I don't see a problem with clock indicator.
image
It looks good to me. Could you explain in details what is a problem?

domkab:
[image: 40f8428b-20e4-4e3f-b1a0-65add088d0f8.png]
hey, as you can see the indicators start offsetting and are not equally
aligned around the circle shadow: especially you can see it at 90 degrees
where top border is longer and white.


also the second arrow starts misaligninging further down the clock.

[image: 2.png]
and if you look at the gaps between indicator they are not even at times,
why so ?


2023-01-03, an, 13:33 Igor Goncharenko ***@***.***> rašė:

qoa11a (approved):
Good job! 👍

PR #1971, opened 2023-01-04
test

vova552:
- [DEMO LINK](https://vova552.github.io/layout_stop-watch/)


src/styles/untils/blocks/stopwatch.scss
```diff
@@ -0,0 +1,39 @@
+.stopwatch {
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
mstrilec, line 4: `align-items` doesn't change anything, so it can be removed

src/styles/untils/blocks/stopwatch.scss
```diff
@@ -0,0 +1,39 @@
...
+  position: relative;
+
+  &__second {
+    position: absolute;
+    top: $mainTop;
+    width: 8px;
+    height: $mainHeight;
```
mstrilec, line 11: these values (in addition to `width`), are used 2 times (in `&__second` and `&__minute`), so they can be exported to the mixin

mstrilec (changes requested):
Well done! This is a good job, but it could be shortened a bit.

mstrilec (approved):
GJ 🔥

PR #1972, opened 2023-01-04
watches was created2

Daria13k:
https://Daria13k.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,25 @@
+.watch {
+  @include watch ($watch-size, $border-color, $background-color-watch);
+  position: relative;
+  transform: rotate(180deg);
+
+  &__hand-minute {
+    @include hand-minute ($width-hand-minute, 
+    $height-hand-minute, $color-hand-minute);
+    position: absolute;
+    top: 50%;
+    left: calc(50% - #{$width-hand-minute} / 2);
+    animation: move steps($step-animation) $time-animation-minute infinite;
+    transform-origin: center 0;
+  }
+
+  &__hand-second {
+    @include hand-second ($width-hand-second, 
+    $height-hand-second, $color-hand-second);
+    position: absolute;
+    top: 50%;
+    left: calc(50% - #{$width-hand-second} / 2);
+    animation: move steps($step-animation) $time-animation-second infinite;
+    transform-origin: center 0;
+  }
+}
--- a/src/styles/utils/_mixins.scss
+++ b/src/styles/utils/_mixins.scss
@@ -1,0 +1,30 @@
+@mixin watch ($watch-size, $border-color, $background-color-watch) {
+  box-sizing: border-box;
+  width: $watch-size;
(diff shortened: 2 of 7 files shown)


src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,30 @@
...
+  border-radius: $width-hand-minute / 2;
+}
+
+@mixin hand-second ($width-hand-second, 
+$height-hand-second, $color-hand-second) {
+  width: $width-hand-second;
+  height: $height-hand-second;
+  background-color: $color-hand-second;
+  border-radius: $width-hand-second / 2;
+}
```
Viktor-Kost, line 24: the same mixins. Use only one for both hands

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,25 @@
...
+
+  &__hand-second {
+    @include hand-second ($width-hand-second, 
+    $height-hand-second, $color-hand-second);
+    position: absolute;
+    top: 50%;
+    left: calc(50% - #{$width-hand-second} / 2);
+    animation: move steps($step-animation) $time-animation-second infinite;
+    transform-origin: center 0;
+  }
```
Viktor-Kost, line 24: a lot of repeatable styles. 
You can just combine like this
```
a,
b {
 {{your_styles}}
}
```

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,30 @@
+@mixin watch ($watch-size, $border-color, $background-color-watch) {
```
Viktor-Kost, line 1: mixin is not about separating styles, it is about reusing styles. 
So this watch mixin is not reusable or flexible
I would name it like `circle`, with only   
```
width: $size;
height: $size;
border-radius: 50%;
```
and maybe ` background-color: $background-color-watch;` but with a default value
Other styles i would set directly

Viktor-Kost (changes requested):
Nice work!

Yet we can apply some improvements

src/styles/utils/_mixins.scss
Daria13k: the variables are different for both hands. how can I use it?

src/styles/blocks/watch.scss
Daria13k: the same, the variables are different for both hands. how can I use it?

src/styles/utils/_mixins.scss
Daria13k: Ok. I will try to change the code. Thank you!

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,35 @@
+.watch {
+  @include circle ($size, $background-color-circle);
+  box-sizing: border-box;
```
OlEzhA0, line 3: add this to `*` selector

OlEzhA0 (approved):
GJ!

PR #1980, opened 2023-01-08
commit message

Peti51:


Peti51:
[DEMO LINK](https://Peti51.github.io/layout_stop-watch/)

src/styles/clock.scss
```diff
@@ -0,0 +1,142 @@
+.clock {
+  position: absolute;
+  height: 200px;
+  width: 200px;
```
polosanya, line 4: image

src/styles/body.scss
```diff
@@ -0,0 +1,10 @@
+* {
+  margin: 0;
+  padding: 0;
```
polosanya, line 3: Do you really need to reset margins and padding for all the elements?
Isn't only body enough?

src/styles/clock.scss
```diff
@@ -0,0 +1,142 @@
...
+  bottom: 0;
+  margin: auto;
+
+  &__top {
+    position: absolute;
+    width: 3px;
+    height: 8px;
+    background-color: #262626;
+    left: 0;
+    right: 0;
```
polosanya, line 20: Pay attention to your CSS properties order
https://9elements.com/css-rule-order/

src/styles/clock.scss
```diff
@@ -0,0 +1,142 @@
...
+    left: 0;
+    bottom: 0;
+    margin: 0 auto;
+  }
+
+  &__left {
+    position: absolute;
+    width: 8px;
+    height: 3px;
```
polosanya, line 49: Don't duplicate yourself, Almost all the styles are the same for all elements. 
Can't you just add the general class for all of them or create a mixin?

src/styles/variables.scss
```diff
@@ -0,0 +1,15 @@
+:root {
+  --white-color: #fff;
+  --clock-big-border-color: #34626c;
+  --background-color: #262626;
+  --bigger-pointer-height: 8px;
+  --smaller-pointer-height: 3px;
+  --bigger-pointer-width: 8px;
```
polosanya, line 7: You can use SASS variables. They don't require root and look cleaner
https://sass-lang.com/documentation/variables

src/styles/body.scss
```diff
@@ -0,0 +1,7 @@
+body {
+  background-color: #839b97;
```
polosanya, line 2: Would be great to create variables for all colors on page

src/styles/clock.scss
```diff
@@ -0,0 +1,148 @@
+.clock {
+  position: var( --position-of-elements);
```
polosanya, line 2: In opposite, I think it's redundant to create variable for position

polosanya (changes requested):
A few small changes

PR #1988, opened 2023-01-12
stopwatch done

kirill-vypirovskyi:
- [DEMO LINK](https://kirill-vypirovskyi.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,28 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <section class="stop-watch">
+      <div class="stop-watch__time stop-watch__time--0">0</div>
+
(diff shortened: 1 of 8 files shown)


src/index.html
```diff
@@ -1,14 +1,28 @@
...
+
+      <div class="stop-watch__time stop-watch__time--30">30</div>
+
+      <div class="stop-watch__time stop-watch__time--45">45</div>
+
+      <div class="stop-watch__seconds"></div>
+
+      <div class="stop-watch__minutes"></div>
+
+      <div class="stop-watch__center"></div>
```
MaksymMaliuk, line 25: i think it's to many blank lines. Also you can youse mix classes for watch-hands like this :
```suggestion
      <div class="stop-watch__hand stop-watch__hand--sec"></div>
      <div class="stop-watch__hand stop-watch__hand--min"></div>
```

src/styles/utils/_animations.scss
```diff
@@ -0,0 +1,19 @@
...
+
+@keyframes minutes-arrow-move {
+  from {
+    transform: translateX(-50%) rotate(getMinutesStartPos($stopwatch-time));
+  }
+
+  to {
+    transform: translateX(-50%) rotate(0deg);
+  }
+}
```
MaksymMaliuk, line 19: You don't need to use two keyframes for same animations.

MaksymMaliuk (changes requested):
You need to make some changes. But it's nice work.

src/styles/utils/_animations.scss
kirill-vypirovskyi: It is the same animation, there are different functions in start rotate position

src/index.html
```diff
@@ -1,14 +1,22 @@
...
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <section class="stop-watch">
+      <div class="stop-watch__time stop-watch__time--0">0</div>
+      <div class="stop-watch__time stop-watch__time--15">15</div>
+      <div class="stop-watch__time stop-watch__time--30">30</div>
+      <div class="stop-watch__time stop-watch__time--45">45</div>
```
Andr1yK, line 16: u can generate content from scss

src/styles/utils/_animations.scss
Andr1yK: agree with @MaksymMaliuk

Andr1yK: u can use css steps

Andr1yK (approved):

PR #2003, opened 2023-01-16
Watch

HannaVasylieva:
- [DEMO LINK](https://HannaVasylieva.github.io/layout_stop-watch/)


src/styles/container.scss
```diff
@@ -0,0 +1,21 @@
...
+$watch-size: 500px;
+$width-hand: 3px;
+
+@mixin round-block ($size) {
+  width: $size;
+  height: $size;
+  border-radius: 50%;
+}
+
+@mixin height-hand ($height, $color) {
```
Viktor-Kost, line 11: I would change this mixin name, because you use not only height inside. It can be confused
Probably, i don't even use mixin for such changes

Viktor-Kost (approved):
Great job!!

PR #2012, opened 2023-01-19
commit message

AlexRaven93:
- [DEMO LINK](https://AlexRaven93.github.io/layout_stop-watch/)


src/styles/watch/watch.scss
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  &__hand-minute {
+    position: relative;
+    top: 130px;
+    width: $hand-width;
+    height: 120px;
+    background-color: #f00;
+    border-radius: 4px;
+    animation: 360s infinite linear turnover;
```
DarkMistyRoom, line 36:
```suggestion
    animation: 3600s infinite linear turnover;
```

src/styles/main.scss
```diff
@@ -0,0 +1,14 @@
...
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
DarkMistyRoom, line 13: better use a separate container for this. it is a good idea to position the body

src/styles/watch/watch.scss
```diff
@@ -0,0 +1,39 @@
+.watch {
+  display: flex;
+  justify-content: center;
+  height: $clock-size;
+  width: $clock-size;
+  margin: 0 auto;
```
DarkMistyRoom, line 6: Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

src/styles/watch/watch.scss
```diff
@@ -0,0 +1,39 @@
...
+  &__hand-minute {
+    position: relative;
+    top: 130px;
+    width: $hand-width;
+    height: 120px;
+    background-color: #f00;
+    border-radius: 4px;
+    animation: 360s infinite linear turnover;
+    transform-origin: center 120px;
+  }
```
DarkMistyRoom, line 38: you have repeated styles here. would be better to move them to extend or common class to avoid repeating

DarkMistyRoom (changes requested):

src/styles/watch/stopwatch.scss
```diff
@@ -0,0 +1,43 @@
+.stopwatch {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
ViktorHavryliuk, line 5: This is relevant to positions too, don’t use it on block itself
image

ViktorHavryliuk (changes requested):

ViktorHavryliuk (approved):

PR #2024, opened 2023-01-24
made watch

IvanVaverchak:
[DEMO LINK](https://IvanVaverchak.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,41 @@
+.watch {
+  height: $watch-size;
+  width: $watch-size;
+  border: 1px solid $minutes-color;
+  border-radius: 50%;
+  display: flex;
+  justify-content: center;
+  position: relative;
+
+  &__minute {
+    height: 250px;
+    width: 15px;
+    border-radius: 8px;
+    background-color: $minutes-color;
+    z-index: 1;
+    transform-origin: center;
+    animation: time minute-rotate($second-time) linear 0s infinite;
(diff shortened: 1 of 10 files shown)


src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,5 @@
+@mixin placing {
```
Viktor-Kost, line 1: this mixin is not flexible and can be used only at some specific situations

src/styles/utils/function.scss
```diff
@@ -0,0 +1,3 @@
+@function minute-rotate($second-time) {
```
Viktor-Kost, line 1: why do you need a word rotate? Use more meaningful names. 
Function has to be started with a verb and describe what is going on inside
Ex: calculate-minutes

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,41 @@
...
+    animation: time $second-time steps(60) 0s infinite;
+    transform-origin: center;
+    @include placing;
+  }
+
+  &::after {
+    content: "";
+    height: 30px;
+    width: 30px;
+    border-radius: 50%;
```
Viktor-Kost, line 36: you have used these combination twice. So make a mixin with a name "circle" which gets a size and create a circle

Viktor-Kost (changes requested):

Viktor-Kost (approved):
GJ!

PR #2030, opened 2023-01-25
Solution for layout_stop-watch

pindel-kateryna:
[DEMO LINK](https://pindel-kateryna.github.io/layout_stop-watch/)


src/index.html
```diff
@@ -1,14 +1,22 @@
...
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="stopwatch">
+      <div class="stopwatch__minutes">
+        <div class="stopwatch__min"></div>
+      </div>
+      
+      <div class="stopwatch__secunds">
```
witflash, line 17: You can name your classes as you want, but `seconds` is the correct word in English :)

src/styles/main.scss
```diff
@@ -0,0 +1,19 @@
+* {
+  margin: 0;
+  padding: 0;
```
witflash, line 3: Move `margin` and `padding` outside of `*` selector. You can use in `*` just `box-sizing` rule. 
Use it for all tags you need:
```
body,
h1,
p,
etc... {
  margin: 0;
  padding: 0;
}
```
But here you need just for `body` I think.

witflash (changes requested):
Great design! I like it!
Could you do small fixes according to my comments, please?

Viktor-Kost (approved):

PR #2034, opened 2023-01-29
Develop

SerhiiNalyvaiko:
- [DEMO LINK](https://SerhiiNalyvaiko.github.io/layout_stop-watch/)


src/styles/stopwatch.scss
```diff
@@ -0,0 +1,73 @@
...
+  position: absolute;
+  top: 0;
+  right: 0;
+  left: 0;
+  bottom: 0;
+  margin: auto;
+  background-color: #ffd700;
+  border: 4px #daa520;
+  border-color: #daa520;
+  box-shadow: 10px;
```
TarasHoliuk, line 11: it doesn't work
image

src/styles/stopwatch.scss
```diff
@@ -0,0 +1,73 @@
...
+  height: 500px;
+  width: 500px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 100%;
+
+  &::before {
+    content: "";
+    position: absolute;
```
TarasHoliuk, line 21: it is centered because of the `flexbox`
So no need to add  `position: absolute;` here. Or vice versa - you can remove `display: flex` from `.stopwatch`

src/styles/stopwatch.scss
```diff
@@ -0,0 +1,73 @@
+.stopwatch {
+  position: absolute;
+  top: 0;
+  right: 0;
+  left: 0;
+  bottom: 0;
```
TarasHoliuk, line 6: Not critical in this task, but if you've added these properties just to center the stopwatch better to add some styles to the container above (`body` in this case) to do it

TarasHoliuk (changes requested):
GJ :+1: Everything is fine with animation, but I've added a few general comments

nazarmatsevych (approved):

PR #2035, opened 2023-01-29
Stop watch develop

misha-vynnyk:
[DEMO LINK](https://misha-vynnyk.github.io/layout_stop-watch/)


src/index.html
```diff
@@ -1,14 +1,20 @@
...
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body class="page__body">
+    <div class="stopwatch stopwatch--position">
+      <div class="stopwatch__minute-hand">
+        <div class="stopwatch__center-hand">
+          <div class="second-hand"></div>
```
polosanya, line 15:
```suggestion
          <div class="stopwatch__second-hand"></div>
```
And do you really need to nest them?

polosanya (changes requested):
Why does the minutes hand start not from 0? Check it by yourself
image

src/index.html
misha-vynnyk: Yes, I think it was right to put the hand in the middle. Since it is part of the center-hand.

VitaliyBondarenko1982 (approved):

PR #2041, opened 2023-01-31
clock is done

vladyslav-mosor:
- [DEMO LINK](https://vladyslav-mosor.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -2,11 +2,7 @@
   @extend %center-object;
   @include circle($clock-size);
 
-  background-image: linear-gradient (
-    109.6deg,
-    rgba(245,56,56,1) 11.2%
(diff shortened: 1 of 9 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,78 @@
+.clock {
+  @extend %center-object;
```
rashyd-hasratov, line 2: 8. [BEM & STYLES] - Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

rashyd-hasratov (changes requested):
Nice! But there is a problem according to BEM.

rashyd-hasratov (approved):

illiarerun (approved):
I really like your solution, GREAT JOB

PR #2097, opened 2023-02-08
stop-watch

SergeyShapovalenko:
- [DEMO LINK](https://SergeyShapovalenko.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/_watch.scss
+++ b/src/styles/blocks/_watch.scss
@@ -1,41 +1,34 @@
-.wrapper {
+.watch {
+  @include circle($watch-size);
   @include flex($position-center);
-
+    
   display: flex;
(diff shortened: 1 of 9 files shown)


src/styles/blocks/_watch.scss
```diff
@@ -0,0 +1,41 @@
+.wrapper {
```
ViktorHavryliuk, line 1: Move wrapper to it's own file, each BEM block should be in separate file, especially as there is no need to nest .watch class inside in this case

ViktorHavryliuk (approved):
Good work, but there's one suggestion, please implement it

DarkMistyRoom (approved):

PR #2104, opened 2023-02-13
Mission comleted! Watch was created

RatuhWD:
- [DEMO LINK](https://ratuhwd.github.io/layout_stop-watch/)


etojeDenys (changes requested):
image

the watch stopped after one spin, the animation should last forever

PR #2105, opened 2023-02-13
create stop-watch

Anastasiia-Shahoval:
[DEMO LINK](https://anastasiia-shahoval.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/animation.scss
+++ b/src/styles/blocks/animation.scss
@@ -1,0 +1,19 @@
+@keyframes move {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(#{$rotation});
+  }
+}
+
+@keyframes run {
+  from {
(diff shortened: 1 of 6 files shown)


src/styles/blocks/animation.scss
```diff
@@ -0,0 +1,19 @@
...
+
+@keyframes run {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(#{$rotation});
+  }
+}
```
OlEzhA0, line 19: why do we need 2 similar keyframes?

OlEzhA0 (changes requested):
would be better to do something like 
https://user-images.githubusercontent.com/60295799/218707803-f10305d0-6087-4e82-b33c-e198271693c5.mov

etojeDenys (approved):
LGTM

PR #2112, opened 2023-02-14
Develop

AnnaNazarova23:
DEMO LINK - https://AnnaNazarova23.github.io/layout_stop-watch/


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,44 @@
...
+  width: $stopwatchSize;
+  height: $stopwatchSize;
+  background-color: rgb(235, 245, 145);
+  border: 2px solid yellow;
+  border-radius: 50%;
+
+
+  &__dot {
+    position: absolute;
+    top: 50%;
```
YegorKorenkov, line 12: also to center the element you need to use transform: translate(-50%, -50%). Because absolute left: 50% places the element exactly in the middle of the parent element, but children ( dot in your case) has its own width, in this case you need to move dot by half the width. The same for height

YegorKorenkov (commented):
Small remark, but everything okey. Nice work

src/styles/utils/_keyframes.scss
```diff
@@ -0,0 +1,19 @@
...
+
+@keyframes secondHand {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+}
```
etojeDenys, line 19: why do you need to write two similar keyframes?

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #2121, opened 2023-02-18
Develop stopwatch

HannieA:
[DEMO LINK] https://HannieA.github.io/layout_stop-watch/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,25 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -1,14 +1,25 @@
...
+  </head>
+  <body>
+    <div class="container">
+      <div class="stopwatch">
+        <div class="stopwatch__hands">
+          <div class="arrows stopwatch__min-arrow"></div>
+        </div>
+        <div class="stopwatch__hands">
+          <div class="arrows stopwatch__sec-arrow"></div>
+        </div>
```
etojeDenys, line 19:
```suggestion
        <div class="stopwatch__hands">
          <div class="arrows stopwatch__min-arrow"></div>
          <div class="arrows stopwatch__sec-arrow"></div>
        </div>
```

maybe we can do smth like that?

etojeDenys (changes requested):
good job

few nits:

let move colors to variables

src/index.html
HannieA: I added wrappers for each hand for correct centre positioning. If I add just 1 wrapper for two hands it will centre this block but not hands separately

YegorKorenkov: Yes, in your case fo correct centering yoy should add wrappers. But, for example, you can remove wrappers and add width/height of arrows to variable and after calculate top/left position calc(50% - $min-height). But your implementation looks more cleaner

PR #2122, opened 2023-02-19
Was created stopwatch with buttons, use pseudo-class ":checked"

gretoq:


Ambulence (approved):
So...Looking cool! User interaction looks very cool!  To my might just you made some properties hard size, also some properties do not work such as 'align-items' in your container, you can  check, if you remove nothing break

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,131 @@
+.watch {
+  display: flex;
+
+  height: 400px;
+  border: 8px solid $c-dark;
+  border-radius: 10%;
+
+  background: #adadad;
```
sergii-nosachenko, line 8: Nice to have all colors as vars

src/styles/utils/_animation.scss
```diff
@@ -0,0 +1,23 @@
...
+  animation-play-state: running;
+}
+
+#stop:checked ~ * .watch__hand {
+  animation-play-state: paused;
+}
+
+#reset:checked ~ * .watch__hand {
+  animation: none;
+}
```
sergii-nosachenko, line 11: Better to have separate file for `controls`.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,131 @@
+.watch {
+  display: flex;
+
+  height: 400px;
```
sergii-nosachenko, line 4: A lot of hardcoded values. Think about how to decrease the such amount. F.e. calculate them, use flexbox props, use relative units etc.

sergii-nosachenko (approved):
Good work 👍
Always think about code style, reusability, and optimization.

PR #2125, opened 2023-02-19
stop-watch OkM

Oksana-Maruniak:
[DEMO LINK](https://Oksana-Maruniak.github.io/layout_stop-watch/)


src/index.html
```diff
@@ -1,14 +1,25 @@
...
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <main class="page">
+      <section class="page__clock clock">
+        <img 
+        class="clock__image"
+        src="./images/PinClipart.com_golf-clipart_279835.webp" 
+        alt="mickey mouse on the clock"
+        >
```
ViktorHavryliuk, line 18:
```suggestion
        <img 
          class="clock__image"
          src="./images/PinClipart.com_golf-clipart_279835.webp" 
          alt="mickey mouse on the clock"
        >
```

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,64 @@
...
+    @include transform;
+    width: 6px;
+    height: 140px;
+    border-radius: 10px;
+    background-color: $red-color;
+    animation: move 15s steps(60) infinite forwards;
+    transform-origin: top left;
+  }
+
+  @keyframes move {
```
ViktorHavryliuk, line 46: move keframes to separate file

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,64 @@
...
+    from {
+      transform: rotate(180deg);
+    }
+
+    to {
+      transform: rotate(540deg);
```
ViktorHavryliuk, line 52: Better to set from 0 to 360 deg and set starting position so that arrow points up at 0 degrees

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  &__min {
+    @include position;
+    @include transform;
+    width: 6px;
+    height: 180px;
+    border-radius: 10px;
+    background-color: #000;
+    animation: move 180s steps(12) infinite forwards;
```
ViktorHavryliuk, line 31: Please use real time here for both second and minutes hands

ViktorHavryliuk (changes requested):

src/styles/blocks/clock.scss
Oksana-Maruniak: I don't know how I can do it. I did use rotate: 180deg; for min and sec hand... I think it's incorrect, because during animation active state everything looks right. But when I stop animation in dev tools - hands position go down and they go go beyond the parent element.

vlad-demchuk (approved):

PR #2134, opened 2023-02-23
The markup of the stopwatch is created according to the task

artem-kyrylenko:
- [DEMO LINK](https://artem-kyrylenko.github.io/layout_stop-watch/)


liliia-livandovskaya (approved):
Awesome 👍

PR #2139, opened 2023-02-25
stopwatch created

ielizarov:
- [DEMO LINK](https://ielizarov.github.io/layout_stop-watch/)


src/styles/blocks/page.scss
```diff
@@ -0,0 +1,16 @@
...
+.page {
+  &__body {
+    margin: 0;
+  }
+
+  &__stopwatch {
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%,-50%);
```
lerastarynets, line 14: as i said before, no need to use both top/left and translate

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,46 @@
...
+  &__minute-hand {
+    position: absolute;
+    top: calc(50% - #{$length-minute-hand});
+    left: calc(50% - 0.5 * #{$width-minute-hand});
+    width: $width-minute-hand;
+    height: $length-minute-hand;
+    background-color: #0f0;
+    border-radius: 50%;
+    transform-origin: bottom center;
+    animation: ticker 3600s steps(60) 0.5s infinite;
```
lerastarynets, line 34: u r reusing animation property, I think u can create a mixin in separate file which would look something like following
```
@mixin ticker ($period, $position) {
  animation: rotate #{$period} infinite linear;
  transform-origin: #{$position};

  @keyframes rotate {
    from {
      transform: rotate(0);
    }
    to {
      transform: rotate(360deg);
    }
  }
}
```
and just include the mixin instead of `animation` property in your classes

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

src/styles/blocks/stopwatch.scss
ielizarov: Done. Moved animation to mixin. In the mixin there is only one parameter - period. 
All other parameters need to stay always the same.

src/styles/blocks/page.scss
ielizarov: Top, left - takes parent sizes, transform: translate - takes block sizes. Works fine when body height = 0px .
Flexbox method "align-items: center" only works when I set body "height: 100vh", otherwise it fails.
flex-align-justify

loralevitska (changes requested):
Almost done)

src/styles/main.scss
```diff
@@ -0,0 +1,4 @@
+@import "./utils/variables";
+@import "./utils/mixins";
```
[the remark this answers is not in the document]
ielizarov: Thank you! mixins.scss - added. 
But I had some troubles when was trying to "git push origin develop" to the same commit.
 if simple push don't work,  is it OK to use force push like this:
  git add .
  git commit -a --amend  ( I want to use the same commit)
  git push -f origin develop

maxim2310 (approved):
Good work!!!

PR #2141, opened 2023-02-26
Develop

YuliaYanovskaya:
[DEMO LINK](https://YuliaYanovskaya.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,68 @@
...
+    width: 3px;
+    background-color: #3cda0c;
+    border-radius: 3px;
+    animation: rotate 3600s steps(60) infinite;
+  }
+
+  &__second {
+    @include position-transform-hands;
+    top: -8%;
+    height: 350px;
```
TarasHoliuk, line 52: a bit strange that the arrow is longer than the stopwatch radius:
image

src/styles/main.scss
```diff
@@ -0,0 +1,68 @@
...
+  position: absolute;
+  left: $hands-indent;
+  transform-origin: bottom;
+}
+
+body {
+  margin: 0;
+}
+
+.stop-watch {
```
TarasHoliuk, line 22: According to BEM it's better to move this block to its own file
And the same for mixins and variables

TarasHoliuk (changes requested):

PR #2145, opened 2023-02-28
initial commit

grinjk1993:
https://grinjk1993.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/blocks/container.scss
+++ b/src/styles/blocks/container.scss
@@ -1,0 +1,8 @@
+.container {
+  position: relative;
+  top: 50%;
+  transform: translateY(50%);
+  width: 500px;
+  height: 500px;
+  margin: 0 auto;
+}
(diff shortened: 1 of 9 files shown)


src/styles/blocks/container.scss
```diff
@@ -0,0 +1,8 @@
+.container {
+  position: relative;
+  top: 50%;
+  transform: translateY(50%);
+  width: 500px;
+  height: 500px;
+  margin: 0 auto;
```
DarkMistyRoom, line 7: Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

src/styles/blocks/seconds.scss
```diff
@@ -0,0 +1,9 @@
+.seconds {
+  @include hand;
+  @include position;
```
DarkMistyRoom, line 3: Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

DarkMistyRoom (changes requested):
use BEM. you can create `.clock` and `.clock__minutes` and `.clock__seconds` for example

IvanFesenko (approved):

PR #2146, opened 2023-02-28
animated clock with stopped animation on hover

DenysKrysanov:
- [DEMO LINK](https://DenysKrysanov.github.io/layout_stop-watch/)


src/styles/block/clock.scss
```diff
@@ -0,0 +1,54 @@
+@mixin handSize {
+  position: absolute;
+  bottom: 50%;
+  left: 40%;
+  border-radius: 10px;
+  width: 5px;
+}
```
igorkruz, line 7: you should create _mixin.scss file for this code

src/styles/block/clock.scss
```diff
@@ -0,0 +1,54 @@
...
+.clock {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+
+  &__center {
+    background-color: black;
+    height: 20px;
+    width: 20px;
+    border-radius: 50%;
```
igorkruz, line 18: here it would be best if you used z-index to cover the arrows

src/styles/block/clock.scss
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @keyframes spin {
+    100% {
+      transform: rotate(360deg);
+    }
+  }
```
igorkruz, line 25: also, you can create an animation.scss file to this code

igorkruz (approved):
I guess with these suggestions your code would be better

src/styles/block/clock.scss
liliia-livandovskaya: Agree, let's move this block to `_mixin.scss` file

src/styles/block/clock.scss
liliia-livandovskaya: add `position: relative;` and `z-index: 1`. Arrows should be under the clock`s center

src/styles/block/clock.scss
liliia-livandovskaya: Agree, it would be better to create an `animation.scss` file

liliia-livandovskaya (changes requested):
Great job! Pay attention to the comments

liliia-livandovskaya (approved):

PR #2155, opened 2023-03-01
Create stopwatch

AnyaLyaLya:
- [DEMO LINK](https://AnyaLyaLya.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/stop-watch.scss
+++ b/src/styles/blocks/stop-watch.scss
@@ -1,0 +1,78 @@
+.stop-watch {
+  display: block;
+
+  padding: 50px;
+
+  background: #eae7dc;
+
+  &__box {
+    position: relative;
+
+    margin: auto;
+
+    width: $stop-watch-box;
+    height: $stop-watch-box;
+
+    background-color: #eae7dc;
+    border: 15px dotted #e85a4f;
+    border-radius: 50%;
+  }
+
+  &__center {
+    box-sizing: border-box;
+
+    position: absolute;
+    z-index: 2;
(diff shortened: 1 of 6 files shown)


src/styles/blocks/stop-watch.scss
```diff
@@ -0,0 +1,78 @@
...
+
+    width: $stop-watch-minute;
+    height: 225px;
+
+    background-color: #8e8d8a;
+    border-radius: 80%;
+
+    animation-name: time;
+    animation-duration: 60s * 60;
+  }
```
Kavon-Skela, line 50: I think it will be better to add an animation-timing-function with the value "linear". It will make a second arrow to move such as real second arrow

Kavon-Skela (approved):
Try to look at the second arrow throughout the animation. When it comes to the end of the animation, it starts to move slower

src/styles/blocks/stop-watch.scss
```diff
@@ -0,0 +1,78 @@
+.stop-watch {
+  display: block;
+
+  padding: 50px;
```
SerhiiKirik, line 4:
```suggestion
  height: 100vh;
  padding: 50px;
```

src/styles/blocks/stop-watch.scss
```diff
@@ -0,0 +1,78 @@
...
+
+  background: #eae7dc;
```
SerhiiKirik, line 6: don't remember to create variables for colors

```suggestion
  background: $c-color;
```

SerhiiKirik (approved):
Good job 🔥

PR #2161, opened 2023-03-01
Develop

roshirin:
[DEMO LINK](https://roshirin.github.io/layout_stop-watch/)

It was too late when I noticed the png demo. My watch is electronic with digitals on it. But it still uses animation effects and demanded more complicated logic to implement. So please check it out, maybe it'll fit the task.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # Stopwatch task
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_stop-watch/)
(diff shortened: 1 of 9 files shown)


Anna-Kuzmenko:
Very interesting clock!

Anna-Kuzmenko (approved):

liliia-livandovskaya (changes requested):
Sorry, but this solution cannot be accepted. You can make a simple watch with just 2 arrows or just skip this task because it is not required. Next time first read carefully the requirements and only then start working on the task.

VinnieJ-2k20 (approved):

PR #2176, opened 2023-03-03
stopwatch

edgarlokys:
https://edgarlokys.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/blocks/point.scss
+++ b/src/styles/blocks/point.scss
@@ -1,0 +1,9 @@
+.stopwatch__oclock-point {
+  position: relative;
+  left: 49%;
+  top: 50%;
+  @include point-size;
+  border-radius: 50%;
+  background-color: $second-color;
+  transform: translateY(-50%);
+}
--- a/src/styles/utils/_keyframes.scss
+++ b/src/styles/utils/_keyframes.scss
@@ -1,0 +1,19 @@
+@keyframes minute-hand {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+}
+
+@keyframes second-hand {
+  from {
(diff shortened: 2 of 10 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,21 @@
+/*===Colors===*/
```
OleziO, line 1: Remove all comments from the code.

OleziO (changes requested):
Avoid vertical scroll, and align the clock on the page.
image

src/styles/utils/_keyframes.scss
```diff
@@ -0,0 +1,19 @@
...
+
+@keyframes second-hand {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+}
```
etojeDenys, line 19: why do we need to the same `keyframes`
combine it into one

src/styles/blocks/point.scss
```diff
@@ -0,0 +1,10 @@
+.stopwatch__oclock-point {
+  position: relative;
+  overflow: $hidden;
+  left: 49%;
```
etojeDenys, line 4: can we make `left:50%` and just add `transoform: translate(..)`?

etojeDenys (changes requested):

src/styles/utils/_mixin.scss
```diff
@@ -0,0 +1,15 @@
+@mixin size {
+  width: 500px;
+  height: 500px;
+}
```
OlEzhA0, line 4: mb make sense to create some reusable mixin where you can pass size and create square

OlEzhA0 (approved):

PR #2185, opened 2023-03-05
FirstVersion

AlexRog01:
- [DEMO LINK](https://AlexRog01.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/_utils.scss
+++ b/src/styles/_utils.scss
@@ -1,0 +1,21 @@
+@mixin square-size ($width) {
+  width: $width;
+  height: $width;
+}
+
+@mixin arrow-animation($time) {
+  animation-name: move;
+  animation-duration: $time * 1s;
+  animation-timing-function: steps($time, end);
+  animation-iteration-count: infinite;
+  animation-play-state: paused;
+}
+
+%clock-hand-start-position {
+  position: absolute;
+  left: 50%;
+  top: 50%;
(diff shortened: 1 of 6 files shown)


src/styles/_utils.scss
```diff
@@ -0,0 +1,21 @@
+@mixin square-size ($width) {
+  width: $width;
+  height: $width;
+}
+
+@mixin arrow-animation($time) {
+  animation-name: move;
+  animation-duration: $time * 1s;
```
liliia-livandovskaya, line 8: `* 1s` rewrite it without `*`

src/styles/clock.scss
```diff
@@ -0,0 +1,61 @@
...
+    background-color: $main-theme-color;
+    
+    @extend %clock-hand-start-position;
+    @include arrow-animation(3600);
+  }
+  
+  &__center {
+    @include square_size(5%);
+    position: absolute;
+    left: 50%;
```
liliia-livandovskaya, line 30:
```suggestion
    left: 50%;
```
50% can be moved to variables

liliia-livandovskaya (approved):
Great job 👍

src/styles/_utils.scss
AlexRog01: I don’t know how else to indicate that these are seconds, because I just pass a number to the mixin

PR #2195, opened 2023-03-08
create animated stop-watch

AsyaYeromina:
[DEMO LINK](https://AsyaYeromina.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/utils/mixins.scss
+++ b/src/styles/utils/mixins.scss
@@ -1,8 +1,8 @@
 @mixin hands-styles($hand-height, $hand-width) {
   position: absolute;
-  top: setHandTopPosition($hand-height);
+  top: 50%;
   left: 50%;
-  transform: translateX(-50%);
+  transform: translate(-50%, -100%);
   height: $hand-height;
   width: $hand-width;
   transform-origin: bottom center;
--- a/src/styles/utils/animations.scss
+++ b/src/styles/utils/animations.scss
@@ -1,11 +1,11 @@
 @keyframes rotateHands {
 
   0% {
-    transform: translateX(-50%) rotate(0deg);
+    transform: translate(-50%, -100%) rotate(0deg);
   }
 
   100% {
-    transform: translateX(-50%) rotate(360deg);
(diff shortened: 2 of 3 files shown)


src/styles/blocks/stop_watch.scss
```diff
@@ -0,0 +1,44 @@
...
+    position: absolute;
+    top: setHandTopPosition($second-hand-height);
+    left: 50%;
+    height: $second-hand-height;
+    width: 2px;
+    background-color: #000;
+    border-radius: 2px;
+    animation: rotateHands 60s steps(60, end) infinite;
+    transform-origin: bottom center;
+  }
```
slavanosatskyi, line 30: there are repeatable styles like position, left, transform-origin and background-color. Try to come up with solution, when there is no repeatable styles.

slavanosatskyi (changes requested):
nice work, i see only one place to be improved

src/styles/blocks/stop_watch.scss
AsyaYeromina: Done

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,11 @@
+@mixin hands-styles($hand-height, $hand-width) {
+  position: absolute;
+  top: setHandTopPosition($hand-height);
+  left: 50%;
+  transform: translateX(-50%);
```
etojeDenys, line 5: what do you think?
```suggestion
  top: 50%;
  left: 50%;
  transform: translateX(-50%, -50%);
```

etojeDenys (changes requested):
almost done

src/styles/utils/mixins.scss
AsyaYeromina: But I also have another question:
The animation description also has a transform property (rotate). And it seems that it rewrites other transform property (translate)  in the element's styles. So the rotation looks proper only if I duplicate  `translate` in animation. Is that the way it should be? Or there is another decision?

IvanFesenko (approved):
Well done

PR #2198, opened 2023-03-09
task complete

DanielSorokowski:
- [DEMO LINK](https://DanielSorokowski.github.io/layout_stop-watch/)

About checkpoint 3 - I used specific values on clock hands becouse of my design, the images i used are not symetric so if i wanted to make it perfect i needed to adjust the position. With normal shapes there would be no problem


solaryasha (approved):
Good job!

PR #2201, opened 2023-03-09
add clockface

1v43k:
- [DEMO LINK](https://1v43k.github.io/layout_stop-watch/)


src/styles/clock.scss
```diff
@@ -0,0 +1,106 @@
...
+    left: $size-face / 2;
+    height: $seconds-height;
+    width: $seconds-width;
+    background-color: $black-color;
+    border-radius: 30%;
+    transform-origin: bottom;
+    animation-name: seconds;
+    animation-duration: 60s;
+    animation-timing-function: steps(60);
+    animation-iteration-count: infinite;
```
vlad-demchuk, line 38: use animation shortend css property
image

vlad-demchuk (approved):

PR #2203, opened 2023-03-10
stopwatch commit

vladimirshub:
- [DEMO LINK](https://vladimirshub.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/clock.scss
+++ b/src/styles/clock.scss
@@ -1,0 +1,29 @@
+html {
+  background: #282828;
+  text-align: center;
+  font-size: 10px;
+}
+
+body {
+  margin: 0;
+  font-size: 2rem;
+  display: flex;
+  flex: 1;
+  min-height: 100vh;
+  align-items: center;
+}
+
+.clock {
+  background: #FFF;
+  border: 7px solid #282828;
+  box-shadow: -4px -4px 10px rgba(67,67,67,0.5),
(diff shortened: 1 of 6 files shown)


src/styles/clock_hands.scss
```diff
@@ -0,0 +1,25 @@
+.minute {
+  height: 100px;
+  width: 4px;
+  background: #777;
+  position: absolute;
+  left: 50%;
+  top: 40px;
+  animation: tick 3600s infinite linear;
+  -webkit-animation: tick 3600s infinite linear;
```
Volodymyr-Melnyk-Fron10der, line 9: move all magic numbers into variables

src/styles/clock.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+body {
+  margin: 0;
+  font-size: 2rem;
+  display: flex;
+  flex: 1;
+  min-height: 100vh;
+  align-items: center;
+}
```
Volodymyr-Melnyk-Fron10der, line 14: 1. move to reset.scss
2. there are a lot styles for body, leave only margin, anf move others into separated container 
3. smae for html,` text-align: center;` looks like doesn't belong there

Volodymyr-Melnyk-Fron10der (changes requested):
Good job🔥 Keep it up)

PR #2207, opened 2023-03-11
clock

artyomwhite:
- [DEMO LINK](https://artyomwhite.github.io/layout_stop-watch/)


src/styles/clock.scss
```diff
@@ -0,0 +1,62 @@
...
+    position: absolute;
+    top: calc(50% - 3px);
+    left: 50%;
+    height: 10px;
+    width: 140px;
+    background: black;
+    border-radius: 5px;
+    animation: animate 3600s linear infinite;
+    transform-origin: left;
+  }
```
VitaliyBondarenko1982, line 38: before and after have almost same styles.
make mixin for them with necessary parameters

VitaliyBondarenko1982 (changes requested):

src/styles/clock.scss
artyomwhite: tell me why this mixin does not work in a separate file?! if for example move "./block/clock.scss" and mixin "./utils/mixin.scss".

artyomwhite: i include @import "./block/clock.scss"; and @import "./utils/mixin.scss"; file stops working!

artyomwhite: good?

src/styles/block/clock.scss
```diff
@@ -0,0 +1,44 @@
...
+
+@keyframes animate {
+  0% {
+    transform: rotate(-90deg);
+  }
+
+  100% {
+    transform: rotate(270deg);
+  }
+}
```
etojeDenys, line 44: also, you can move keyframes into another file

etojeDenys (approved):
looks good

PR #2208, opened 2023-03-11
Clock animations

Zibi95:
[DEMO LINK](https://Zibi95.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/wrapper.scss
+++ b/src/styles/blocks/wrapper.scss
@@ -1,0 +1,5 @@
+.wrapper {
+  @include flex-center;
+  width: 100vw;
+  height: 100vh;
+}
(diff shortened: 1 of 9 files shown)


src/styles/blocks/wrapper.scss
```diff
@@ -0,0 +1,5 @@
+.wrapper {
+  @include flex-center;
+  width: 100vw;
```
Radoslaw-Czerniawski, line 3: 100vh is fine however don't use 100vw, if your site had a vertical scroll then this would cause a horizontal overflow / scroll. You can check how it behaves by setting on any sites wrapper `width: 100vw;` in the devtools in your browser (even here on github)

Radoslaw-Czerniawski (approved):
One small comment with info for the future. Generally good pr, keep up with the good work :D

PR #2214, opened 2023-03-12
Clock

Ilia991:
https://Ilia991.github.io/layout_stop-watch/


tetiana-ivanchenko (changes requested):
your arrows should be in the center
image

PR #2230, opened 2023-03-16
add task solution

KielpinskiJ:
- [DEMO LINK](https://KielpinskiJ.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/content.scss
+++ b/src/styles/blocks/content.scss
@@ -1,0 +1,6 @@
+.content {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100vh;
+}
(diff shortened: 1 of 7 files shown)


src/styles/blocks/content.scss
```diff
@@ -0,0 +1,6 @@
+.content {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 100vh;
```
DanielSorokowski, line 5: better is to use min-height, instead of height bcs your content can become bigger than 100vh, so its expands,
but that is not a problem with the task so Good job.

DanielSorokowski (approved):

d-wasilewski (approved):
👏

src/styles/blocks/content.scss
d-wasilewski: In general you're right, but it won't happen in this case - he took care of it 🥇

PR #2237, opened 2023-03-17
task completed

Ozwiena:
- [DEMO LINK](https://ozwiena.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/clock.scss
+++ b/src/styles/clock.scss
@@ -1,0 +1,63 @@
+.clock {
+  position: relative;
+  width: $clock-size;
+  height: $clock-size;
+  border-radius: 50%;
+  border: 6px solid burlywood;
+  background-color: beige;
+
+  &__hours {
+    position: absolute;
+    top: 300px;
+    left: 50%;
+    height: $hour-hand-length;
+    width: 8px;
+    border-radius: 4%;
+    background-color: saddlebrown;
+
+    transform-origin: top center;
+
+    animation-name: tickingHours;
+    animation-duration: 3600s;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+  }
+
+  &__minutes {
+    position: absolute;
+    top: 300px;
+    left: 50%;
+    height: $minute-hand-length;
+    width: 8px;
+    border-radius: 4%;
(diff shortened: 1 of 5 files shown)


src/styles/clock.scss
```diff
@@ -0,0 +1,53 @@
...
+  position: relative;
+  width: $clock-size;
+  height: $clock-size;
+  border-radius: 50%;
+  border: 6px solid burlywood;
+  background-color: beige;
+
+  &__hours {
+    position: absolute;
+    top: 300px;
```
sylcym, line 11: instead of `top: 300px` you can write `top: 50%`, this way if you change the container height it will be always centered

src/styles/clock.scss
```diff
@@ -0,0 +1,53 @@
...
+
+    animation-name: ticking;
+    animation-duration: 3600s;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+  }
+
+  &__minutes {
+    position: absolute;
+    top: 300px;
```
sylcym, line 28: same as before :)

src/styles/main.scss
```diff
@@ -0,0 +1,10 @@
+@import "_variables.scss";
+@import "clock.scss";
+
+body {
+  margin: 0;
+
+  .clock {
+    margin: auto;
+  }
+}
```
sylcym, line 10: i think you can move `.clock` selector to the `clock.scss` file, and the `body { margin: 0; }` code to the `_reset.scss` file :)

sylcym (changes requested):

src/styles/main.scss
Ozwiena: I did not want to use the margin property for a BEM block, that is why I put .clock selector setting the element's margins within the 'body' block.

sylcym (approved):

src/styles/clock.scss
michalcywka: +1

src/styles/clock.scss
michalcywka: +1

michalcywka (approved):

PR #2246, opened 2023-03-19
Stopwatch task

diana-tuz:
[DEMO LINK](https://diana-tuz.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,61 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stopwatch {
+  margin: 50px auto;
+  border: 2px solid #030;
+  border-radius: 50%;
+  width: 600px;
+  height: 600px;
+  position: relative;
+
+  &__second {
+  width: 200px;
+  height: 10px;
+  background-color: #030;
+  border-radius: 20px;
+  position: absolute;
+  top: 50%;
+  left: 67%;
+  transform-origin: left center;
+  animation: rotate 60s linear infinite;
+}
+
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,61 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+}
+
+.stopwatch {
```
vlad-demchuk, line 9: check this checklist point
image

src/styles/main.scss
```diff
@@ -0,0 +1,61 @@
...
+  margin: 50px auto;
+  border: 2px solid #030;
+  border-radius: 50%;
+  width: 600px;
+  height: 600px;
+  position: relative;
```
vlad-demchuk, line 15: change css properties order according to these rules

vlad-demchuk: check all such cases

src/styles/main.scss
```diff
@@ -0,0 +1,61 @@
...
+      height: 25px;
+      background-color: #003;
+      border-radius: 50%;
+      position: absolute;
+      bottom: -5px;
+      left: -8px;
+    }
+  }
+
+  @keyframes rotate {
```
vlad-demchuk, line 53: separate file for keyframes

vlad-demchuk (changes requested):
don't forget to use sass variables
image

OleziO (changes requested):
The arrow of seconds must be larger than the minute's arrow.

slavanosatskyi (approved):

PR #2251, opened 2023-03-20
task done

artem-ivashchenko:
- [DEMO LINK](https://artem-ivashchenko.github.io/layout_stop-watch/)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,14 @@
+$clock-size: 500px;
+$clock-radius: $clock-size / 2;
+$center-width: $clock-size / 25;
+
+$minute: 120s;
```
polosanya, line 5: Are you sure that the minute hand makes a full revolution in 120 seconds?

polosanya (changes requested):
Pay attention to the minute hand. It moves too fast
image

src/styles/utils/_variables.scss
artem-ivashchenko: just wanted to show that it is moving, i know that minute hand makes a full rotation during 3600s

OleziO (approved):

PR #2252, opened 2023-03-21
Solution

OlhaKissa:
[DEMO LINK](https://olhakissa.github.io/layout_stop-watch/)


src/index.html
```diff
@@ -1,14 +1,26 @@
...
+      <div class="minutes"></div>
+    </div>
+
+    <div class="centre-container">
+      <div class="centre"></div>
+    </div>
+
+    <div class="second-container">
+      <div class="seconds"></div>
+    </div>
```
IvanFesenko, line 23: As I can see, no need to create 3 different classes `minute-container`, `centre-container`, `second-container` it could be one class

IvanFesenko (changes requested):
Push all files to github

src/styles/main.scss
```diff
@@ -0,0 +1,7 @@
+@import "./blocks/reset.scss";
+
+@import "./blocks/centre.scss";
+@import "./blocks/minutes.scss";
+@import "./blocks/seconds.scss";
+
+@import "./utils/keyframes";
```
IvanFesenko, line 7: Where I can see those files?

nazarmatsevych: Looks like you didn't add those files to commit, try to add them

nazarmatsevych (changes requested):
Hey, good work, 
Try to fix all of the previous comments before making re-request

If you have any questions - feel free to ask them in fe_chat

PR #2261, opened 2023-03-23
My Stop-Watch

SashaViun:
- [DEMO LINK](https://SashaViun.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,38 @@
+.watch {
+  @include clock;
+  background-color: black;
+  width: $watch-size;
+  height: $watch-size;
+  margin-top: 50px;
+  border: 4px solid $main-color;
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  cursor: pointer;
+
+  &__center {
+    @include design;
+    @include clock;
+    height: $center-size;
+    width: $center-size;
+  }
+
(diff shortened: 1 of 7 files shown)


src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
+@keyframes minutes {
+  100% {
```
YegorKorenkov, line 2: Please and 0% point for animation, because arrows very slow down in the interval of 50-60 seconds

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,15 @@
...
+    transform: rotate(360deg);
+  }
+}
+
+@keyframes seconds {
```
YegorKorenkov, line 7: Animation are same. You can use only one keyframes for every animation

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,38 @@
...
+
+  &__center {
+    @include design;
+    @include clock;
+    height: $center-size;
+    width: $center-size;
+  }
+
+  &__seconds {
+    @include design;
```
YegorKorenkov, line 22: design mixin and position mixin both have a position:absolute rule, try to not repeat it

YegorKorenkov (changes requested):
Looks good, please check remarks and fix

PR #2262, opened 2023-03-23
added task solution

mariuszgit:
https://mariuszgit.github.io/layout_stop-watch/


src/styles/blocks/_page.scss
```diff
@@ -0,0 +1,7 @@
+.page {
+  align-items: center;
+  display: flex;
+  height: 100vh;
+  justify-content: center;
+  width: 100%;
+}
```
DorotaLeniecDev, line 7: Styles are good here, it is also a good practice to sort css properties so the code is cleaner and easier to ready. Also you dont need to set 100% width. By default, the <body> element in HTML will take up the entire width of the viewport, which is the visible area of the web page in the browser window. The default width of block-level elements in HTML is 100% of the parent container's width
```suggestion
.page {
  display: flex;
  justify-content: center;
  align-items: center;
  height: 100vh;
}
```

src/styles/blocks/_stop-watch.scss
```diff
@@ -0,0 +1,72 @@
...
+
+  &::after {
+    content: "";
+    display: block;
+    width: 20px;
+    height: 20px;
+    position: absolute;
+    left: 50%;
+    top: 50%;
+    // border: 1px solid black;
```
DorotaLeniecDev, line 33: Remove comment

DorotaLeniecDev (changes requested):
Good job, I have left few comments :)

DorotaLeniecDev (approved):
Well done :)

PR #2265, opened 2023-03-24
solution

DaniilBarilotti:
https://daniilbarilotti.github.io/layout_stop-watch/index.html


src/styles/main.scss
```diff
@@ -0,0 +1,62 @@
...
+    background-color: #124197;
+    border-radius: 20px;
+    position: absolute;
+    bottom: 50%;
+    left: 241px;
+    transform-origin: bottom center;
+    animation: move-seconds 60s linear infinite;
+  }
+  
+  @keyframes move-minutes {
```
mykhalenych, line 43: you can use one keyframes but add different animation-duration
for example for seconds 
`animation-duration: 60s;`
for minutes:
`animation-duration: 3600s;`

mykhalenych (changes requested):

PR #2272, opened 2023-03-26
add task solution

ArtemBehter:


Radoslaw-Czerniawski (approved):
Wow, ok this is a very nice looking clock. :) You could try tinkering with scss to reuse some the styling logic, for example you could try using `@mixin`:
```
@mixin clock-hands ($z-index, $height, $color, $duration) {
  position: absolute;
  bottom: 50%;
  z-index: $z-index;
  height: $height;
  background-color: $color;
  animation: clock $duration infinite;
}
```
Something like that, however the pr is good and I approve it :)

PR #2275, opened 2023-03-28
stopwatch

oleksandr-rudiuk:


src/styles/animation.scss
```diff
@@ -0,0 +1,21 @@
...
+    transform: rotate(90deg);
+  }
+
+  50% {
+    transform: rotate(180deg);
+  }
+
+  75% {
+    transform: rotate(270deg);
+  }
```
DarkMistyRoom, line 16: why do you need all these options? it would be the same result if you left 0% and 100%

src/styles/watch.scss
```diff
@@ -0,0 +1,41 @@
+.watch {
+  position: relative;
+  @include square($watch-size);
+  border: 1px black solid;
+  margin-left: #{$watch-shift};
```
DarkMistyRoom, line 5: Don't add external styles (positioning or margins) to BEM-blocks. Use mix where necessary and move all external styles under element selector.

src/styles/watch.scss
```diff
@@ -0,0 +1,41 @@
...
+    top: #{$minutes-top};
+    left: #{$hands-left};
+    height: #{$height-minutes};
+    width: #{$hands-width};
+    background-color: red;
+    border-radius: 5px;
+    margin: 0;
+    transform-origin: bottom center;
+    animation: time 60s linear 1;
+  }
```
DarkMistyRoom, line 40: is it modifications of a watch? look more like elements of it

DarkMistyRoom, line 40: try to avoid duplicating. use extends or common classes

DarkMistyRoom (changes requested):
Good, but let's make your clock more like usual clocks) with minute and second hands) And let's center it on vertical

src/styles/watch.scss
oleksandr-rudiuk: There isn't duplicating. Minutes and seconds hands have different sizes and shifts.

PR #2286, opened 2023-04-02
stop-watch1

viktor-kniahnytskyi:
- [DEMO LINK](https://viktor-kniahnytskyi.github.io/layout_stop-watch/)


src/styles/clock.scss
```diff
@@ -0,0 +1,54 @@
...
+}
+
+@keyframes rotate-second {
+  from {
+    transform: rotate(0deg) translateX(-50%) translateY(-100%);
+  }
+  to {
+    transform: rotate(360deg) translateX(-50%) translateY(-100%);
+  }
+}
```
etojeDenys, line 54: move to another file
and why do we need two the same keyframes

etojeDenys (changes requested):

src/styles/clock.scss
viktor-kniahnytskyi: > move to another file and why do we need two the same keyframes

i used two for minutes and seconds

src/styles/keyframes.scss
```diff
@@ -0,0 +1,17 @@
...
+}
+
+@keyframes rotate-second {
+  from {
+    transform: rotate(0deg) translateX(-50%) translateY(-100%);
+  }
+  to {
+    transform: rotate(360deg) translateX(-50%) translateY(-100%);
+  }
+}
```
VitaliyBondarenko1982, line 17: not fixed. no need two same keyframes

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #2310, opened 2023-04-13
create stopwatch

olha-rypich:
[DEMO LINK](https://olha-rypich.github.io/layout_stop-watch/)


VitaliyBondarenko1982 (changes requested):
Nice watch.
minute-hand and second-hand has almost same styles.
consider to make mixin for them with arguments which different.

DarkMistyRoom (approved):
Pretty nice)

PR #2314, opened 2023-04-13
Solution v.1

annperetiatko:
[DEMO LINK]
https://annperetiatko.github.io/layout_stop-watch/


src/styles/main.scss
```diff
@@ -0,0 +1,42 @@
+@import "keyframes";
+
+$borderRadius: 8px;
+$handWidth: 100px;
+
+.stopwatch {
```
vlad-demchuk, line 6: separate file per bem block

vlad-demchuk: image

vlad-demchuk (changes requested):

src/styles/main.scss
sTorba24: still not fixed

sTorba24 (changes requested):
Fix previous comment please

src/styles/main.scss
```diff
@@ -0,0 +1,38 @@
+@import "keyframes";
+@import "variables";
+@import "mixins";
+
+.stopwatch {
```
sTorba24, line 5: move it to stopwatch.scss and import in main

sTorba24 (changes requested):
Clarified what needs to be done

slavanosatskyi (approved):

PR #2321, opened 2023-04-17
clock created

pr1syazhniy:
[DEMO LINK](https://pr1syazhniy.github.io/layout_stop-watch/)


src/styles/stopwatch.scss
```diff
@@ -0,0 +1,36 @@
+.clock {
+  width: $clock-size;
+  height: $clock-size;
+
+  @include position-value;
+  position: absolute;
```
vlad-demchuk, line 6: image

vlad-demchuk (changes requested):
it would be better to create separate folders for blocks and sass utils
image

src/styles/stopwatch.scss
pr1syazhniy: yeah,forgot about that)

vlad-demchuk: not fixed https://github.com/mate-academy/layout_stop-watch/pull/2321/files#r1169569301
you should move external styles to bem element. you can use mix(refer to checklist examples)

vlad-demchuk (changes requested):

src/index.html
```diff
@@ -1,14 +1,22 @@
...
+          initial-scale=1.0,
+          maximum-scale=1.0,
+          minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="container">
+      <div class="clock">
```
vlad-demchuk, line 16: you have used the bad example)
image

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,38 @@
+.container {
+  @include position-value;
+  position: absolute;
+}
```
vlad-demchuk, line 4: follow checklist 
image

make parent element of clock a bem block and make clock a bem element
don't forget to creaet a separate file for created bem block

vlad-demchuk (changes requested):

PR #2340, opened 2023-04-26
add task solution

Bonum-taurus:
https://Bonum-taurus.github.io/layout_stop-watch


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,43 @@
+.stopwatch {
+  margin: 0 auto;
```
vlad-demchuk, line 2: image
checklist

vlad-demchuk (changes requested):

VitaliyBondarenko1982 (approved):

PR #2349, opened 2023-04-27
task solution

Yuliya1907:
[DEMO LINK](https://Yuliya1907.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,31 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+      content="width=device-width,
+      user-scalable=no, initial-scale=1.0,
+      maximum-scale=1.0, minimum-scale=1.0"
+    >
+    <meta
+      http-equiv="X-UA-Compatible"
+      content="ie=edge"
+    >
+    <title>Stop watch</title>
+    <link
+      rel="stylesheet"
+      href="./styles/main.scss"
+    >
+  </head>
+  <body>
+    <main class="page">
+      <section class="watch">
+        <div class="watch__big-circle">
+          <div class="watch__small-circle"></div>
+          <div class="watch__minutes"></div>
+          <div class="watch__seconds"></div>
+        </div>
+      </section>
+    </main>
+  </body>
 </html>
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,64 @@
+.watch {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 400px;
+  height: 400px;
+
+  &__big-circle {
+    position: relative;
+    width: 300px;
+    height: 300px;
+    border-radius: 50%;
+    border: 4px solid $-main-color;
+    background-color: rgb(245, 245, 168);
+  }
+
+  &__small-circle {
+    position: absolute;
+    top: $-elements-margin;
+    right:$-elements-margin;
+    width: 15px;
+    height: 15px;
+    z-index: 1;
+
+    background-color: $-main-color;
+    border-radius: 50%;
+  }
+
+  &__minutes {
+    position: absolute;
+    left: $-elements-margin + 4;
+    top: 10px;
+    width: 4px;
+    height: 130px;
+
+    background-color: $-main-color;
+    animation: roundMinutes 60s infinite;
+    transform-origin: bottom;
+
+    @keyframes roundMinutes {
+      to {
+        transform: rotate(6deg);
+      }
+    }
+  }
+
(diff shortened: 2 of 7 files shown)


src/index.html
```diff
@@ -1,14 +1,31 @@
...
+  </head>
+  <body>
+    <main class="page">
+      <section class="watch">
+        <div class="watch__big-circle">
+          <div class="watch__small-circle"></div>
+          <div class="watch__minutes"></div>
+          <div class="watch__seconds"></div>
+        </div>
+      </section>
```
julshvets, line 28: No need to create extra wrappers. You can write styles from a class 'watch__big-circle' to class "watch" and delete 1 div.
And check p.1 of the checklist (Don't forget about semantics, div development is not a better option).

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,64 @@
...
+    height: 130px;
+
+    background-color: $-main-color;
+    animation: roundMinutes 60s infinite;
+    transform-origin: bottom;
+
+    @keyframes roundMinutes {
+      to {
+        transform: rotate(6deg);
+      }
```
julshvets, line 43: I think you should add from{...} to @keyframes. even if it is 0. And you can use @keyframes roundSeconds for both animations. You have just to set different times of animation.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,64 @@
...
+    height: 145px;
+
+    background-color: orange;
+    animation: roundSeconds 60s infinite linear;
+    transform-origin: bottom;
+
+    @keyframes roundSeconds {
+      to {
+        transform: rotate(360deg);
+      }
```
julshvets, line 61: I think you should add from{...} to https://github.com/Keyframes. even if it is 0.

julshvets (approved):
Hello! Great job! 
But I have some recommendations.

src/styles/blocks/watch.scss
Ihor-Karpyn: For what?

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,64 @@
+.watch {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 400px;
+  height: 400px;
```
Ihor-Karpyn, line 6: mixin square will be useful here

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,64 @@
...
+
+  &__big-circle {
+    position: relative;
+    width: 300px;
+    height: 300px;
+    border-radius: 50%;
```
Ihor-Karpyn, line 12: Mixin circle will be useful here

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,64 @@
...
+  &__small-circle {
+    position: absolute;
+    top: $-elements-margin;
+    right:$-elements-margin;
+    width: 15px;
+    height: 15px;
+    z-index: 1;
+
+    background-color: $-main-color;
+    border-radius: 50%;
```
Ihor-Karpyn, line 26: circle again

src/styles/main.scss
```diff
@@ -0,0 +1,10 @@
+@import "utils/variables.scss";
+@import "./blocks/watch";
+@import "utils/normalize.css";
+@import "utils/reset.css";
+
+.page {
```
Ihor-Karpyn, line 6: It is recommended to have a separate file for each BEM block.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,2 @@
+$-elements-margin: 140px;
```
Ihor-Karpyn, line 1: It would be beneficial to transfer the variables that are solely connected to one block to the file associated with that specific block. Not clear to which element does the variable refer

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,2 @@
...
+$-main-color: rgb(97, 37, 97);
```
Ihor-Karpyn, line 2: Let's start var names from later

Ihor-Karpyn (changes requested):
Almost done

src/styles/utils/mixins.scss
```diff
@@ -1,11 +1,13 @@
...
-  width: $small-circle-size;
-  height: $small-circle-size;
+@mixin bigcircle ($size) {
+  @include square ($size);
   border-radius: 50%;
 }
+
+@mixin smallcircle ($size) {
+  @include bigcircle ($size * 0.05);
+}
```
Ihor-Karpyn: Please combine the mixins into one and specify the size difference when passing it to the new mixin.

Ihor-Karpyn (changes requested):

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,61 @@
+$elements-margin: 140px;
+
+.watch {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  @include square;
```
Ihor-Karpyn, line 7: I'm not certain if it works without specifying the size. Have you tested it since making the changes? Please check if it's functioning properly and deploy the updated code to gh-pages.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,61 @@
...
+
+  &__big-circle {
+    position: relative;
+    @include circle;
```
Ihor-Karpyn, line 11: Here the same

Ihor-Karpyn (changes requested):

Ihor-Karpyn (approved):
GJ 🔥

PR #2358, opened 2023-04-29
add task solution

Liliya14:
- [DEMO LINK](https://Liliya14.github.io/layout_stop-watch/)


src/styles/utils/_keyframes.scss
```diff
@@ -0,0 +1,9 @@
+@keyframes arrowMove {
+  from {
+    transform: translate(-50%, -100%) rotate(0deg);
```
YegorKorenkov, line 3: Translate redundant if it does not change in any way during the animation

YegorKorenkov (approved):

PR #2361, opened 2023-04-30
add task solution

snuber-w:
- [DEMO LINK](https://snuber-w.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,79 @@
+.watch {
+  position: relative;
+  @include watch-size;
+
+  &__bg {
+    position: absolute;
+    @include watch-size;
+
+    background-color: blueviolet;
+
+    border: 45px solid burlywood;
+    border-radius: 50%;
+
+  }
+
+  &__seconds {
+    position: absolute;
+    @include second-size;
+    z-index: 3;
+
+    border-radius: $second-size / 5;
+    background-color: bisque;
+    border: 1px solid blanchedalmond;
+    left: $watch-size / 2;
+    top: $watch-size / 2;
+
+    animation-name: second-rotation;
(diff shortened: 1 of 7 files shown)


ruslanpashkov (changes requested):
what if we want to change the size of our stopwatch? let's make it more liquid
image

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,79 @@
...
+    background-color: blue;
+    border: 1px solid blue;
+    left: $watch-size / 2;
+    top: $watch-size / 2;
+
+    animation-name: second-rotation;
+    animation-duration: 60s * 60;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+    transform-origin: $minute-width/4 3px;
```
YeMyr, line 49: it is a repeating animation script that could be combined into one common class, so the only difference between seconds and minutes is the duration, which can be specified by a modifier

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,79 @@
...
+    background-color: black;
+    border: 1px solid black;
+    left: $watch-size / 2;
+    top: $watch-size / 2;
+
+    animation-name: second-rotation;
+    animation-duration: 60s * 60 * 60;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+    transform-origin: $hour-width/4 3px;
```
YeMyr, line 67: the same story: repeating animation script

YeMyr (approved):
Hi, great job! Sorry, I left some comments. 
Please, take into consideration comment regarding mixins file.

naziamoff (changes requested):
please check the comments of your colleagues, a lot of valuable points there

PR #2365, opened 2023-05-01
(ym) stop watch

YeMyr:
[DEMO LINK](https://yemyr.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,64 @@
+.watch {
+  @include page-center;
+  @include watch-size;
+  border: 5px solid gray;
+  border-radius: 50%;
+
+  &__hand {
+    @include page-center;
+    animation-name: hands-movement;
+    animation-iteration-count: infinite;
+    animation-timing-function: steps(60, end);
+    transform-origin: center left;
+
+    @keyframes hands-movement {
+      0% {
+        transform: translate(0, -50%) rotate(-90deg);
+      }
+
+      100% {
+        transform: translate(0, -50%) rotate(270deg);
+      }
+    }
+
+    &--minute {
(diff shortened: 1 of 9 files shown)


src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,64 @@
...
+  @include page-center;
+  @include watch-size;
+  border: 5px solid gray;
+  border-radius: 50%;
+
+  &__hand {
+    @include page-center;
+    animation-name: hands-movement;
+    animation-iteration-count: infinite;
+    animation-timing-function: steps(60, end);
```
naziamoff, line 11: can we just combine them into a single property?

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    &--minute {
+      @include minute-hand-size;
+      background-color: $main-color;
+      animation-duration: 3600s;
+    }
+
+    &--second {
+      @include second-hand-size;
```
naziamoff, line 31: If that's not what you mean - you want it to be called like `watch-hand-minutes` or sth 😄 

image

naziamoff (approved):
great job! See one small suggestion

src/styles/blocks/watch.scss
YeMyr: 😁

src/styles/blocks/watch.scss
YeMyr: I'm sure we can, but actualy I don't get how to scpecify animation-duration in this case? 

animation: hands-movement ```?``` steps(60, end) infinite;

So as, I use one common animation property for both of hands and separated class modifiers to apply duration...
Can you advise how it should be write down?

PR #2375, opened 2023-05-03
Develop

drobnayana:
[DEMO LINK](https://drobnayana.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -0,0 +1,11 @@
+@import "../styles/utilis/__reset.scss";
+@import "../styles/utilis/mixins.scss";
+@import "../styles/block/watch.scss";
+
+body {
+  background-color: rgb(229, 205, 241);
```
OleziO, line 6: Try to use HEX instead of RGB

OleziO (approved):
Very nice work! Make the seconds arrow longer than minutes.

PR #2384, opened 2023-05-08
Clock

BeginerAlex:
- [DEMO LINK](https://begineralex.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/clock.scss
+++ b/src/styles/clock.scss
@@ -1,0 +1,126 @@
+.main {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-height: 100vh;
+  width: 100%;
+  background-color: #123141;
+}
+
+.clock {
+  width: $clock-size;
+  height: $clock-size;
+  margin: 65px auto;
+  position: relative;
+  display: flex;
+  border: 30px solid $clock-color;
+  border-radius: 50%;
+  box-shadow: $clock-shadow;
+
+  &:before {
+    content: "";
+    width: $clock-point-size;
+    height: $clock-point-size;
+    position: relative;
+    margin: auto;
+    background: $clock-color;
+    border-radius: 50%;
+    z-index: 11;
+  }
+
+  &__hours {
+    position: absolute;
+    top: $clock-hand-position;
+    left: $clock-hand-position;
+    width: $clock-hand-size;
+    height: $clock-hand-size;
+    display: flex;
+    justify-content: center;
+    border-radius: 50%;
+    transform: rotate(180deg);
+    animation: rotateHours 43200s linear infinite;
+
+    &::before {
+      content: "";
+      width: 8px;
+      height: 120px;
+      background-color: #ff105e;
+      z-index: 10;
+      border-radius: $clock-hand-border;
+      box-shadow: $clock-hand-hour-shadow;
+    }
+  }
+
+  &__minutes {
+    position: absolute;
+    top: $clock-hand-position;
+    left: $clock-hand-position;
+    width: $clock-hand-size;
+    height: $clock-hand-size;
+    display: flex;
+    justify-content: center;
+    border-radius: 50%;
+    transform: rotate(180deg);
+    animation: rotateSeconds 3600s linear infinite;
+
+    &::before {
+      content: "";
+      width: 6px;
+      height: 190px;
+      background-color: #fcf9fa;
+      z-index: 10;
+      border-radius: $clock-hand-border;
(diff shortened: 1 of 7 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,4 @@
+@import "variables";
+@import "reset";
+@import "normalaize";
```
sTorba24, line 3: There is a typo in the import statement. It should be 'normalize' instead of 'normalaize'.

src/styles/reset.scss
```diff
@@ -0,0 +1,6 @@
+*,
+*::after,
+*::before {
+  box-sizing: border-box;
+  font-family: "Roboto", Arial, Helvetica, sans-serif;
```
sTorba24, line 5: It is not recommended to set the font-family in a global reset file. Consider moving it to a more specific selector or create a base styling file for typography.

src/styles/variables.scss
```diff
@@ -0,0 +1,31 @@
...
+  0 0 75px #ff105e;
+$clock-hand-minute-shadow: 0 0 10px #fff,
+  0 0 20px #fff,
+  0 0 30px #fff,
+  0 0 40px #0ff,
+  0 0 70px #0ff,
+  0 0 80px #0ff,
+  0 0 100px #0ff,
+  0 0 150px #0ff;
+$clock-hand-secunde-shadow: 0 0 10px #1bcdc4,
```
sTorba24, line 19: Typo in variable name: Change $clock-hand-secunde-shadow to $clock-hand-second-shadow

src/styles/normalaize.scss
```diff
@@ -0,0 +1,9 @@
+html,
```
sTorba24, line 1: The file name 'normalaize.scss' has a typo. It should be 'normalize.scss'.

src/styles/clock.scss
```diff
@@ -0,0 +1,126 @@
...
+    position: absolute;
+    top: $clock-hand-position;
+    left: $clock-hand-position;
+    width: $clock-hand-size;
+    height: $clock-hand-size;
+    display: flex;
+    justify-content: center;
+    border-radius: 50%;
+    transform: rotate(180deg);
+    animation: rotateSeconds 3600s linear infinite;
```
sTorba24, line 64: The animation for the minute hand is incorrect. Change 'rotateSeconds' to 'rotateMinutes' to use the correct animation.

src/styles/reset.scss
sTorba24: html or body, for example

src/styles/clock.scss
```diff
@@ -0,0 +1,126 @@
...
+      box-shadow: $clock-hand-hour-shadow;
+    }
+  }
+
+  &__minutes {
+    position: absolute;
+    top: $clock-hand-position;
+    left: $clock-hand-position;
+    width: $clock-hand-size;
+    height: $clock-hand-size;
```
sTorba24, line 59: you repeated it 3 times, so i think you can avoid it adding common class to clock elements

sTorba24 (changes requested):
Good work! Lets improve it

src/styles/normalize.scss
```diff
@@ -0,0 +1,10 @@
+html,
+body,
+img,
+p,
+a,
+div {
+  margin: 0;
+  padding: 0;
+  font-family: "Roboto", Arial, Helvetica, sans-serif;
```
TarasHoliuk, line 9: again, font-family should be set for the entire page (e.g html or body tags) and probably inside another file (for example a separate file `fonts`)

src/styles/reset.scss
```diff
@@ -0,0 +1,5 @@
+*,
```
TarasHoliuk, line 1: I'd say it might be in the normalize.scss as well
It's better to create a separate file for fonts (see comment above)

TarasHoliuk (changes requested):
GJ :+1: And amazing styles - look :fire: 
I've added 2 comments - the main thing is to move `font-family` out from `div` selector

src/styles/clock.scss
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  @keyframes rotateSeconds {
+    from {
+      transform: rotate(180deg);
+    }
+    to {
+      transform: rotate(540deg);
+    }
```
etojeDenys, line 100: why do we need three same keyframes

etojeDenys (changes requested):

PR #2388, opened 2023-05-09
Develop

ruslan-ihnatenko:
- [DEMO LINK](https://ruslan-ihnatenko.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/utils/_resets.scss
+++ b/src/styles/utils/_resets.scss
@@ -1,0 +1,6 @@
+body,
+html {
+  margin: 0;
+  padding: 0;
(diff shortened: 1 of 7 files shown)


src/styles/utils/_resets.scss
```diff
@@ -0,0 +1,6 @@
+body,
+html {
+  margin: 0;
+  padding: 0;
+  box-sizing: border-box;
```
witflash, line 5: This rule you should use for `*` selector

witflash (approved):
Well done 👍🏻

PR #2391, opened 2023-05-10
add task layout_stop-watch

P-Nazar:
- [DEMO LINK](https://P-Nazar.github.io/layout_stop-watch/)


src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,86 @@
+.watch {
+  position: relative;
+  top: 100px;
```
etojeDenys, line 3: The watch is not centered vertically. Use flexbox to center the watch both vertically and horizontally.

etojeDenys (approved):

PR #2394, opened 2023-05-11
finished task

yanatoto:
- [DEMO LINK](https://yanatoto.github.io/layout_stop-watch/)


src/index.html
```diff
@@ -10,5 +10,10 @@
 </head>
 <body>
   <h1>Stop watch</h1>
+  <div class="watch">
+    <div class="watch__center"></div>
+    
+    
+  </div>
```
polosanya, line 17:
```suggestion
  <div class="watch">
    <div class="watch__center"></div>
  </div>
```
Don't leave empty lines

src/styles/main.scss
```diff
@@ -0,0 +1,70 @@
+.watch {
+  position: relative;
+  width: 500px;
+  height: 500px;
```
polosanya, line 4: It would be great to create variables for all main values
image

src/styles/main.scss
```diff
@@ -0,0 +1,70 @@
...
+
+  &__center {
+    width: 30px;
+    height: 30px;
+    background-color: black;
+    border-radius: 50%;
+    z-index: 20;
+  }
+
+  &::before {
```
polosanya, line 21: I think its better to add minute and second hands as HTML elements

polosanya (changes requested):
Fix all eslint problems
image

SanyaBratashchuk (approved):

PR #2395, opened 2023-05-11
stop-watch

Marinakyrychynska:
- [DEMO LINK](https://Marinakyrychynska.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/arrows.scss
+++ b/src/styles/blocks/arrows.scss
@@ -1,0 +1,17 @@
+.arrow {
+  position: absolute;
+  left: 50%;
+  bottom: 50%;
+  border-radius: 20px;
+  transform-origin: bottom;
+}
+
+@keyframes moveArrow {
+  from {
+    transform: rotate(0deg);
+  }
+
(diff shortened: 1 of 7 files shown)


src/styles/blocks/arrows.scss
```diff
@@ -0,0 +1,17 @@
+.arrow {
```
polosanya, line 1: It's not critical, but It would be great to name file just like the BEM block: `arrow.scss` instead of `arrowS.scss`

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,30 @@
...
+    height: 48%;
+    background-color: $color-sec;
+    animation: moveArrow 60s infinite steps(60);
+  }
+
+  &__min {
+    width: 12px;
+    height: 40%;
+    background-color: $color-min;
+    animation: moveArrow 3600s infinite steps(60);
```
polosanya, line 19: Let's make animation duration a variable

PR #2397, opened 2023-05-11
Add solution

Roman-Bilichenko:
[DEMO LINK](https://<Roman-Bilichenko>.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,64 @@
+.clock {
+  @include position;
+
+  transform: translate(-50%, -50%);
+  width: 450px;
+  height: 450px;
+  border: 10px double black;
+  border-radius: 50%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  &__center {
+    @include position;
+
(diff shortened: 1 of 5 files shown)


polosanya (changes requested):
Nice UI! Fix the animation's duration according to the task description ⌚ 
Stop watch should have minute and seconds hands. Minute hand makes full rotation in 1 hour, seconds - in 1 minute
image
Also, it would be great to use scss variables for main values

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,64 @@
...
+  @keyframes rotate-minute {
+    0% {
+      transform: translate(-50%, -100%) rotate(0deg);
+    }
+    100% {
+      transform: translate(-50%, -100%) rotate(360deg);
+    }
+  }
+
+  @keyframes rotate-second {
```
YegorKorenkov, line 56: why do we need two same keyframes? please use one

YegorKorenkov (changes requested):

OleziO (approved):

PR #2401, opened 2023-05-12
add task solution

FIREDAN99:
https://FIREDAN99.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,58 @@
+.stopwatch {
+  position: absolute;
+  top: 50%;
+  left: 50%;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+@keyframes second-hand {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
```
etojeDenys, line 57: why do we need two same keyframes

YegorKorenkov (approved):
Looks good, please check remark

PR #2416, opened 2023-05-15
add task solution

monikaprochniewicz:
- [DEMO LINK](https://monikaprochniewicz.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,71 @@
+@import "../utils/variables";
+@import "../utils/mixins";
+
+.clock {
+  border-radius: $circle;
+  @include element-size ($clock-size, $clock-size);
+  border: 6px dotted black;
+  position: absolute;
+  top: $base-pos-value;
+  left: $base-pos-value;
+  transform: translate(-$base-pos-value,
+  (-$base-pos-value));
+
+  &__center {
+    @include default-set;
+    @include element-size ($center-hight,
+    $center-width);
+    @include position ($base-pos-value - ($center-width / 2),
+    $base-pos-value - $center-hight/ 2);
+    border-radius: $circle;
+    z-index: 1;
+    background-color: $center-color;
+    }
+
+  &__hours-arrow {
+    @include default-set;
+    @include element-size ($hours-arrow-height,
+    $hours-arrow-width);
+    @include position ($base-pos-value - ($hours-arrow-width / 2) ,
+    $base-pos-value - $hours-arrow-height);
+    background-color: $hours-arrow-color;
+     @include anim-setting($hours-time);
+    }
+
+  &__minutes-arrow {
+    @include default-set;
+    @include element-size ($minutes-arrow-height,
+    $minutes-arrow-width);
+    @include position ($base-pos-value - ($minutes-arrow-width / 2),
+    $base-pos-value - $minutes-arrow-height );
+    background-color: $minutes-arrow-color;
+    @include anim-setting($minutes-time);
+
+    &::after {
+      position: absolute;
+      content: " ";
+      background-color: $minutes-arrow-color;
+      @include element-size($dot-size, $dot-size);
+      border-radius: $circle;
+      @include position (2 * $base-pos-value , 2 * $base-pos-value);
+      transform: translate(-$base-pos-value,
+      -$base-pos-value);
+    }
+  }
+
+  @keyframes arrow {
+    from {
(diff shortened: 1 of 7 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,71 @@
...
+    &::after {
+      position: absolute;
+      content: " ";
+      background-color: $minutes-arrow-color;
+      @include element-size($dot-size, $dot-size);
+      border-radius: $circle;
+      @include position (2 * $base-pos-value , 2 * $base-pos-value);
+      transform: translate(-$base-pos-value,
+      -$base-pos-value);
+    }
```
DorotaLeniecDev, line 53: This dot is not visible on demo page, verify styles :)

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,20 @@
...
+@mixin element-size ($height , $width) {
+  height: $height;
+  width: $width;
+}
+
+@mixin position ($left , $top) {
+  left: $left;
+  top: $top;
+}
+@mixin default-set {
```
DorotaLeniecDev, line 12: Add **box-sizing: border-box**; to reset.css and specify elements that need to have this property. 
Then you could add **position: absolute**; directly to elements

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,20 @@
...
+
+@mixin element-size ($height , $width) {
+  height: $height;
+  width: $width;
+}
+
+@mixin position ($left , $top) {
+  left: $left;
+  top: $top;
+}
```
DorotaLeniecDev, line 11: Overuse of mixin is not a good practice, in my opinion your code would be more readable if you could simply add those properties directly to elements

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,15 @@
+$clock-size: 300px;
+$circle: 50%;
+$minutes-arrow-height: 50%;
+$minutes-arrow-width: 2%;
+$minutes-arrow-color: green;
```
DorotaLeniecDev, line 5: This is more readable, when you read css without looking into variables.scss file
```suggestion
$color-green: green;
```

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,15 @@
...
+$base-pos-value: 50%;
+$center-hight: 8%;
+$center-width: 8%;
+$center-color: orange;
```
DorotaLeniecDev, line 9: This is more readable, when you read css without looking into variables.scss file
```suggestion
$color-orange: orange;
```

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,15 @@
...
+$hours-arrow-height: 30%;
+$hours-arrow-width: 4%;
+$hours-arrow-color: blue;
```
DorotaLeniecDev, line 12: This is more readable, when you read css without looking into variables.scss file
```suggestion
$color-blue: blue;
```

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,71 @@
...
+  @keyframes arrow {
+    from {
+      transform: rotate(360deg);
+    }
+
+    to {
+      transform: rotate(720deg);
+    }
+  }
+}
```
DorotaLeniecDev, line 65: For simplicity:
```suggestion
@keyframes arrow {
  from {
    transform: rotate(0deg);
  }
  to {
    transform: rotate(360deg);
  }
}
```

DorotaLeniecDev (changes requested):
Clock is working, but code need some more adjustments :)

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,55 @@
+@import "../utils/variables";
+@import "../utils/mixins";
+
+.clock {
```
Radoslaw-Czerniawski, line 4: Don't put every value into a variable, this reduces the code readability. Generally you should put into css variables only constant values that repeat across multiple files so they can be easily accessed and changed without breaking the consistency of the site / layout. For example things like font sizes or colors most likely need to be the same for a feature or whole app so it's a good idea to put them into variables, but things like specific size of one element don't qualify for a variable.

Radoslaw-Czerniawski (changes requested):
Do some slight refactoring, the clock looks very nice so you're almost done :D

Radoslaw-Czerniawski (approved):
Good job :D

PR #2420, opened 2023-05-17
Stop-watch v.2

Dimoshua:
description:
- [DEMO LINK](https://Dimoshua.github.io/layout_stop-watch/)


Dimoshua:
> Doesn't work as expected second hand should make full turn in 1 minute. Minute hand in 1 hour

PR #2425, opened 2023-05-17
stop-watch

KarolinaWiktoria:
https://KarolinaWiktoria.github.io/layout_stop-watch/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,19 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+
+  <body class="page">
+    <div class="page__stopwatch stopwatch">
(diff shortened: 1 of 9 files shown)


src/index.html
```diff
@@ -1,14 +1,19 @@
...
+    <meta charset="UTF-8">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+
+  <body class="page">
```
Radoslaw-Czerniawski, line 13: Don't add classes to the body tag, you can have only one body tag in your markup, just use body tag in css or use a wrapper for the watch

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,37 @@
+.stopwatch {
+  position: relative;
+  @include circle($clock-size);
+  border: thick double $stopwatch-border-color;
```
Radoslaw-Czerniawski, line 4: You shouldn't name your color variables with word color, this is not descriptive, also don't tie the variable to block specifically, rather use generic names describing the actual color like for example `dark-grey`

src/styles/utils/_normalize.scss
```diff
@@ -0,0 +1,7 @@
+html {
+  height: 100vh;
```
Radoslaw-Czerniawski, line 2: This file should be called reset.scss, normalizing is a process of normalizing the css differences between browsers, here you are resetting the styles for the purpose of your work. Also this specific rule is not a reset but rather a styling. An example of reset would be resetting all margins to 0 or setting box-sizing to border-box, but not declaring html's height

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,12 @@
+$second: 1s;
```
Radoslaw-Czerniawski, line 1: Are these 3 variables used anywhere? If not remove them.

Radoslaw-Czerniawski (changes requested):
The clocks looks very nice, however some changes in the code are needed

src/styles/utils/_normalize.scss
KarolinaWiktoria: Thank you @Radoslaw-Czerniawski for your comments. Already finished with fixing :)

darokrk (approved):
Nice one 🥇

PR #2429, opened 2023-05-18
add task solution

OleksandraKoshyk:
[DEMO LINK](https://OleksandraKoshyk.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/page.scss
+++ b/src/styles/blocks/page.scss
@@ -1,0 +1,11 @@
+.page {
+  font-family: Roboto, sans-serif;
+
+  &__body {
+    margin: 0;
+    display: flex;
+    justify-content: center;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/page.scss
```diff
@@ -0,0 +1,11 @@
+.page {
+  font-family: Roboto, sans-serif;
```
OleziO, line 2: Move to html selector and remove from all other places.

OleziO (changes requested):
The hands of the clock are moving too fast. Long hand - must make one circle in 60 seconds, minute hand - 60 * 60 = 3600 seconds for a complete circle

mykhalenych (approved):

PR #2437, opened 2023-05-23
add task solution

Orchibald:
https://orchibald.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,77 @@
+$diameter: 500px;
+$center-diameter: 30px;
+
+.stop-watch {
+  border: 5px white dotted;
+  background-color: #000;
+  height: $diameter;
+  width: $diameter;
+  border-radius: 50%;
+  position: absolute;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,77 @@
...
+    width: 2px;
+    height: 250px;
+    top: 50%;
+    left: 50%;
+    margin-left: -1px;
+    transform-origin: top;
+    animation-duration: 60s;
+    animation-timing-function: linear;
+    animation-iteration-count: infinite;
+    animation-name: rotateSecondHand;
```
VitaliyBondarenko1982, line 57: minute-hand and second-hand has a lot of same properties. you can make mixin for it with arguments which different.

VitaliyBondarenko1982 (changes requested):
Nice work.
Please check and fix some points from task description:

- Screenshot from 2023-05-23 14-20-45
- The starting position of the hands should be at the top.

PR #2439, opened 2023-05-23
catalog__stop-watch

Pashahu1:
[DEMO LINK](https://Pashahu1.github.io/layout_stop-watch/)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,46 @@
+.container {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  width: 80vmin;
+  height: 80vmin;
```
vlad-demchuk, line 7: why did you use `vmin` for width? can we use px?

src/styles/utilities/_extends.scss
```diff
@@ -0,0 +1,25 @@
+%setting-second {
+  position: absolute;
+  top: -400%;
+  left: 40%;
+  transform: translate(0,-10%);
+  width: 10px;
+  height: 225px;
+  background-color: yellow;
+  transform-origin: center bottom;
+  animation: clock-hand-rotate-second 5s infinite linear;
```
vlad-demchuk, line 10: set animation time according to minutes and hours

vlad-demchuk (changes requested):

src/styles/blocks/clock.scss
Pashahu1: сначала и использовал пиксели но почему-то сменил на 80-vmin

Pashahu1: and for have good semetri

Pashahu1: if I do not have vmin in width, then there is no semetry (

nazarmatsevych (approved):
👍

PR #2444, opened 2023-05-25
add task solution

MariaSnegireva:
- [DEMO LINK](https://MariaSnegireva.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,62 @@
+.clock {
+  @extend %radius;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  width: 660px;
+  height: 660px;
+
+  border: 10px dotted black;
+
+  margin: auto;
+
+  &__field {
+    @extend %radius;
+    width: 50px;
+    height: 50px;
+
+    position: relative;
+
+    background-color: yellow;
+
+  }
+
+  &__minute {
+    @extend %format_time;
+    @extend %radius;
+
+    width: 15px;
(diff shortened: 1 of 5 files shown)


src/styles/utils/_extends.scss
```diff
@@ -0,0 +1,14 @@
+%format_time {
+  position: absolute;
+
+  margin: 50%;
```
volodymyr-soltys97, line 4: Use 'top' and 'left' properties instead of 'margin' to center the element.

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,62 @@
+.clock {
+  @extend %radius;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  width: 660px;
```
volodymyr-soltys97, line 8: Use 'width: 80vmin;' and 'height: 80vmin;' instead of fixed pixel values to make the clock size responsive according to the task description.

volodymyr-soltys97 (changes requested):
Good job 👍 

src/styles/main.scss
```diff
@@ -0,0 +1,3 @@
+@import "./utils/extends";
```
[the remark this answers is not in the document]
MariaSnegireva: I know but when I run npm test lint, have this test err:
image

src/styles/blocks/clock.scss
MariaSnegireva: done

vlad-demchuk (approved):

PR #2466, opened 2023-06-01
Develop

NadiaKovalchuk:
https://NadiaKovalchuk.github.io/layout_stop-watch/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,12 +9,12 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <div class="stopwatch">
+    <article class="stopwatch">
       <div class="stopwatch__border">
         <div class="stopwatch__hand stopwatch__hand--minute"></div>
         <div class="stopwatch__hand stopwatch__hand--second"></div>
         <div class="stopwatch__dot"></div>
       </div>
-    </div>
+    </article>
   </body>
 </html>
--- a/src/styles/block/stopwatch.scss
+++ b/src/styles/block/stopwatch.scss
@@ -23,18 +23,16 @@
 
     &--minute {
       height: 100px;
-      width: 2px;
+      width: 1px;
       transform-origin: top;
-      translate: -1px;
       animation: clock-hand-rotate 60s linear infinite;
       background-color: blue;
     }
 
     &--second {
       height: 50px;
-      width: 4px;
+      width: 2px;
       transform-origin: top;
-      translate: -2px;
       animation: clock-hand-rotate 5s linear infinite;
       background-color: yellow;
     }


src/styles/block/stopwatch.scss
```diff
@@ -0,0 +1,50 @@
...
+  justify-content: center;
+  height: 100vh;
+
+  &__border {
+    position: relative;
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    box-sizing: border-box;
+    width: 200px;
```
juljaVysotska, line 13: **Demo has 80vmin size**

src/index.html
```diff
@@ -1,14 +1,20 @@
...
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="stopwatch">
```
juljaVysotska, line 12: You should use here semantic tags like `main`, `article`, etc.

src/styles/block/stopwatch.scss
```diff
@@ -0,0 +1,50 @@
...
+  &__hand {
+    position: absolute;
+    left: 50%;
+    top: 50%;
+
+    &--minute {
+      height: 100px;
+      width: 2px;
+      transform-origin: top;
+      translate: -1px;
```
juljaVysotska, line 28: You can use another way to positioning this element. Because right now you don't follow checklist. [checklist par. 3](https://github.com/mate-academy/layout_stop-watch/blob/master/checklist.md)

Maybe try to make this transform more adaptive? Like, no matter what width I set, the element was always positioned in the center. Think, how we can do this with transform property.

src/styles/block/stopwatch.scss
```diff
@@ -0,0 +1,50 @@
...
+      transform-origin: top;
+      translate: -2px;
+      animation: clock-hand-rotate 5s linear infinite;
+      background-color: yellow;
+    }
+  }
+
+  &__dot {
+    position: absolute;
+    width: 15px;
```
juljaVysotska, line 45: You set with and height in the same way in two places. Maybe we can create mixin for it?

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,8 @@
+@keyframes clock-hand-rotate {
+  from {
+    transform: rotate(-180deg);
+  }
+  to {
+    transform: rotate(180deg);
```
juljaVysotska, line 6: You can use here turn value for transform: rotate([rotate MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate)) for simplicity.

src/styles/block/stopwatch.scss
NadiaKovalchuk: but as i understand. i can use the sizes that i want to? :)

src/index.html
NadiaKovalchuk: Why do i need to use them instead of div? Thanks)

YegorKorenkov (approved):
Good job, but it is better to pay attention to the comments of mentors, they will not advise you anything bad, they will only advise you how to improve your code😊

PR #2467, opened 2023-06-01
add task solution

IraShtynda:
- [DEMO LINK](https://irashtynda.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,21 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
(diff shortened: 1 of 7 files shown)


src/index.html
```diff
@@ -1,14 +1,21 @@
...
+    <meta charset="UTF-8">
+    <meta name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+
+  <body>
+    <div class="watch">
```
juljaVysotska, line 14: You should use here semantic tags like `main`, `article`, etc.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,37 @@
...
+  position: relative;
+  box-sizing: border-box;
+  @include size($watch-size, $watch-size);
+  margin: $margin-block auto;
+  border: $watch-border solid $main-dark-color;
+  border-radius: 50%;
+
+  &__center {
+    @include positionCenter;
+    transform: translate(-50%, -50%);
```
juljaVysotska, line 11: You use `transform: translate()` everywhere where you use mixin `positionCenter`. You need to remake this mixin in next way - mixin should get parameters for transform property. But these parameters must be optional(default value can be -50%, -50%).

src/styles/utils/animations.scss
```diff
@@ -0,0 +1,9 @@
+@keyframes move-hand {
+  from {
+    transform: translate(-50%, -100%) rotate(0);
```
juljaVysotska, line 3: You can use here `turn` value for `transform: rotate`([rotate MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/transform-function/rotate)) for simplicity.

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix them.

Best regards🌸

juljaVysotska (approved):
Good job!

PR #2473, opened 2023-06-02
add task solution

ydashko:
- [DEMO LINK](https://ydashko.github.io/layout_stop-watch/)


src/styles/utils/placeholder.scss
```diff
@@ -0,0 +1,8 @@
+%clock {
+    position: absolute;
+    animation-name: move;
+    animation-duration: 5s;
+    transform-origin: top center;
+    animation-direction: normal;
+    // animation-iteration-count: infinite;
```
mykhalenych, line 7: delete all comments before commit

src/styles/utils/available.scss
```diff
@@ -0,0 +1 @@
+$shift: 50%;
```
mykhalenych, line 1: why does this file available?
I think it is variable

mykhalenych (changes requested):

src/styles/utils/available.scss
ydashko: Yeah, it should've been variable

mykhalenych (approved):

PR #2476, opened 2023-06-02
add task solution

MariiaMonakhova:
- [DEMO LINK](https://MariiaMonakhova.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/utils/keyframes.scss
+++ b/src/styles/utils/keyframes.scss
@@ -1,0 +1,8 @@
+@keyframes rotate {
+  from {
+    transform: rotate(0);
+  }
+  to {
+    transform: rotate(1turn);
+  }
+}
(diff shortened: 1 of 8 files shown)


src/styles/utils/keyframes.scss
```diff
@@ -0,0 +1,8 @@
+@keyframes rotate {
+  from {
+    transform: rotate(0);
```
juljaVysotska, line 3: Your minutes and hour hands don't center, because you don't take into account the width of the arrows during animation. 
Screenshot 2023-06-02 at 16 57 29

When you set `transform: translate(-50%, -50%);` in mixin `absolutePositioning` **without** animation - everything looks and works well, but when you add animation you rewrite transform property of mixin with keyframes values. So we have that bug.

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix them

Best regards🌸

anstsot (changes requested):
Could you redeploy your demo to see the last changes, please?
For now it has the old styles

anstsot (approved):
Good work, approving:sparkles:

PR #2481, opened 2023-06-02
add task solution

dipinchuk:
- [DEMO LINK](https://dipinchuk.github.io/layout_stop-watch/)


src/styles/utils/_reset.scss
```diff
@@ -0,0 +1,5 @@
+* {
+  margin: 0;
+  padding: 0;
```
anstsot, line 3: do not reset margin and padding for all (`*`) elements. please reset only for needed tags

anstsot (approved):
Good work, approving:sparkles:

PR #2491, opened 2023-06-04
add task solution

Platon-jun:
- [DEMO LINK](https://platon-jun.github.io/layout_stop-watch/)


vlad-demchuk (changes requested):
it should work like a real stopwatch, not like your implementation
https://github.com/mate-academy/layout_stop-watch/assets/102411130/426cfcb0-f71f-42cf-b28b-ef5a40df771e

rotation should be infinite

also, you didn't push all your files
image

anastasiiavorobiova (approved):
Good job

PR #2508, opened 2023-06-12
add task solution

Gnydyn:
[DEMO LINK](https://Gnydyn.github.io/layout_stop-watch/)


src/style.css
```diff
@@ -0,0 +1,55 @@
...
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
+.watch__center {
+  position: absolute;
+  top: auto;
```
volodymyr-soltys97, line 45: You can remove the 'top', 'left', 'right', and 'bottom' properties as they are all set to 'auto', which is the default value.

volodymyr-soltys97 (changes requested):
Good job
Visually, it looks like the arrows are not in the very center, check this additionally, please
image

anastasiiavorobiova (changes requested):
Great start! It should work as in the following video and be similar to a real clock. Now it moves a bit randomly.

https://github.com/mate-academy/layout_stop-watch/assets/135232001/276748e8-21f2-47cc-b942-4eea1ecb6e94

nazarmatsevych (approved):
👍

PR #2511, opened 2023-06-12
Develop

ihodyk:
[DEMO LINK](https://ihodyk.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,54 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+.clock {
+  position: relative;
+  height: 80vmin;
+  width: 80vmin;
+  border-radius: 100%;
+  border: 6px dotted black;
+  margin: auto;
+  margin-top: 50px;
+
+  &__center {
+  height: 30px;
+  width: 30px;
+  background-color: #FFA500;
+  border-radius: 100%;
+  position: relative;
+  margin: auto;
+  top: 50%;
+  }
+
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,54 @@
+* {
+  margin: 0;
```
vlad-demchuk, line 2: do not use `*` selector to reset margins

src/styles/main.scss
```diff
@@ -0,0 +1,54 @@
...
+
+  &__center {
+  height: 30px;
+  width: 30px;
+  background-color: #FFA500;
+  border-radius: 100%;
+  position: relative;
+  margin: auto;
+  top: 50%;
+  }
```
vlad-demchuk, line 23:
```suggestion
  &__center {
    height: 30px;
    width: 30px;
    background-color: #FFA500;
    border-radius: 100%;
    position: relative;
    margin: auto;
    top: 50%;
  }
```

vlad-demchuk: `margin: auto` works only for horizontal margins

vlad-demchuk: do not forget about css properties order

vlad-demchuk (changes requested):
It doesn't look great
center the elements
image

vlad-demchuk (changes requested):
the same, it should work for different widths, height
image

volodymyr-soltys97 (approved):
Great job 👍

PR #2517, opened 2023-06-13
Stopwatch

vladkorobka:
https://vladkorobka.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/animations.scss
+++ b/src/styles/animations.scss
@@ -1,0 +1,9 @@
+@keyframes rotation {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+}
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,64 @@
+@import "variables";
+@import "animations";
+
+* {
+  box-sizing: border-box;
+}
+
+body {
+  display: flex;
+  justify-content: center;
+  min-height: 100vh;
+  margin: 0;
+}
+
+.watch {
+  border: 10px dotted #000;
+  border-radius: 50%;
+  display: flex;
+  justify-content: center;
+  width: 80vmin;
+  height: 80vmin;
+  margin: auto;
+
+  &__center {
+    background-color: orange;
+    border-radius: 50%;
+    width: 50px;
+    height: 50px;
+    margin: auto;
+    z-index: 1;
+  }
+
+  &__arrow {
+    position: absolute;
+    bottom: 50%;
+    transform: rotate(0deg);
+    transform-origin: bottom center;
+    animation-name: rotation;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+
+    &--second {
+      display: flex;
+      justify-content: center;
+      align-items: center;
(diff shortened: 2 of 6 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,64 @@
...
+}
+
+body {
+  display: flex;
+  justify-content: center;
+  min-height: 100vh;
+  margin: 0;
+}
+
+.watch {
```
vlad-demchuk, line 15: separate file per bem block

src/styles/main.scss
```diff
@@ -0,0 +1,64 @@
...
+  border: 10px dotted #000;
+  border-radius: 50%;
+  display: flex;
+  justify-content: center;
+  width: 80vmin;
+  height: 80vmin;
+  margin: auto;
```
vlad-demchuk, line 22: css properties order

vlad-demchuk (changes requested):
this clock hand seems too long
image

src/styles/variables.scss
```diff
@@ -0,0 +1,2 @@
+$speed-short: 5s;
```
volodymyr-soltys97, line 1: It is a good practice to add a brief comment explaining the purpose of the variables

src/styles/animations.scss
```diff
@@ -0,0 +1,9 @@
+@keyframes rotation {
```
volodymyr-soltys97, line 1: Consider adding separate animations for minute and second hands to control their rotation speeds independently

volodymyr-soltys97, line 1: Add a more descriptive name for the animation, such as 'rotateHands'

src/styles/main.scss
```diff
@@ -0,0 +1,4 @@
+@import "variables";
```
volodymyr-soltys97, line 1: Consider checking if the imported variables file contains CSS variables for stopwatch, such as colors, sizes, and durations.

src/styles/watch.scss
```diff
@@ -0,0 +1,51 @@
+.watch {
+  position: relative;
+  display: flex;
+  justify-content: center;
+  width: 80vmin;
```
volodymyr-soltys97, line 5: Use 'vw' or 'vh' instead of 'vmin' for width and height to ensure proper scaling on different screen sizes.

src/styles/watch.scss
```diff
@@ -0,0 +1,51 @@
...
+  height: 80vmin;
+  margin: auto;
+  border: 10px dotted #000;
+  border-radius: 50%;
+
+  &__center {
```
volodymyr-soltys97, line 11: For consistency, use the same units for width and height. Change '5vh' to '5%' or '5vw'.

volodymyr-soltys97 (changes requested):
Good job 👍 

src/styles/animations.scss
vladkorobka: thank you for this comment, but I've created this animation for 2 stopwatch hands, I know, that hands will rotate from 0deg to 360deg. Would it make sense to leave it like it is, because this elements do same moves?

PR #2518, opened 2023-06-13
add task solution

Luk2asz:
- [DEMO LINK](https://Luk2asz.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,26 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="container">
+      <div class="watch">
+        <div class="second-axis">
+          <div class="second-axis__hand-watch-second"></div>
+        </div>
+
+        <div class="watch__minute-axis minute-axis">
+          <div class="minute-axis__hand-watch-minute"></div>
+        </div>
+
+        <div class="watch__center watch-center"></div>
+      </div>
+    </div>
+  </body>
 </html>
--- a/src/styles/blocks/minute-axis.scss
+++ b/src/styles/blocks/minute-axis.scss
@@ -1,0 +1,28 @@
+.minute-axis {
+  height: 15px;
+  width: $watch-size * 0.5;
+  display: flex;
+  justify-content: flex-end;
+
+  &__hand-watch-minute {
+    height: 15px;
+    width: $watch-size * 0.25;
+    background-color: red;
+  }
+
+  animation-duration: 3600s;
+  animation-timing-function: linear;
+  animation-delay: 0s;
+  animation-iteration-count: infinite;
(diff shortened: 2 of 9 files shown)


src/index.html
```diff
@@ -1,14 +1,26 @@
...
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="container">
+      <div class="watch">
+        <div class="second-axis">
+          <div class="second-axis__hand-watch-second"></div>
```
darokrk, line 15: The class name 'second-axis__hand-watch-second' should be more descriptive and follow BEM naming convention, consider changing it to 'second-axis__hand'.

src/index.html
```diff
@@ -1,14 +1,26 @@
...
+        </div>
+
+        <div class="watch__minute-axis minute-axis">
+          <div class="minute-axis__hand-watch-minute"></div>
```
darokrk, line 19: The class name 'minute-axis__hand-watch-minute' should be more descriptive and follow BEM naming convention, consider changing it to 'minute-axis__hand'.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,20 @@
...
+
+  &__center {
+    align-self: center;
+    position: absolute;
+  }
+
+  &__minute-axis {
+    align-self: center;
+    position: absolute;
+    justify-content: center;
```
darokrk, line 18: think this property `justify-content: center;` is redundant here if you using an `position: absolute`

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,20 @@
...
+  align-self: center;
+  display: flex;
+  justify-content: center;
+
+  &__center {
+    align-self: center;
+    position: absolute;
+  }
+
+  &__minute-axis {
```
darokrk, line 15: think this whole class should be moved to `minute-axis.scss` and following to BEM convention used as `.minute-axis--center` because it's just a modifier for centering an element

src/styles/blocks/watch-center.scss
```diff
@@ -0,0 +1,7 @@
+.watch-center {
```
darokrk, line 1: Think you don't need this separate file and move it to the watch.scss file. Then this should rather be a block class like `watch__center`

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,20 @@
+.watch {
+  height: $watch-size;
+  width: $watch-size;
+  border: 5px dotted black;
+  border-radius: 50%;
+  align-self: center;
+  display: flex;
+  justify-content: center;
+
+  &__center {
```
darokrk, line 10: this should be a modifier class for `watch__center` something like `watch__center--alignment`

darokrk (changes requested):
UI looks fine, but needs improvements with the code.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,22 @@
...
+    height: $watch-size * 0.03;
+    width: $watch-size * 0.03;
+    background-color: black;
+    border: 5px dotted black;
+    border-radius: 50%;
+  }
+
+  &__center--alignment {
```
darokrk, line 18: When you move this style into just watch__center it won't be enough ? Think you don't need this class anymore :)

src/styles/blocks/minute-axis.scss
```diff
@@ -0,0 +1,30 @@
...
+  position: absolute;
+
+  &__hand {
+    height: 15px;
+    width: $watch-size * 0.25;
+    background-color: red;
+  }
+
+  animation-duration: 3600s;
+  animation-timing-function: linear;
```
darokrk, line 16: Anyway those values, except `animation-duration` are the same as in `second-axis.scss`, you can use a `@mixin` or `@extend` for this to not duplicate code. Example:

Create class e.g name`.hand-animation`, and you can use this class name in HTML `class="second-axis__hand hand-animation"` and `class="minute-axis__hand hand-animation"`

Or you can try to use them here after `animation-duration` as `@extend .hand-animation`, same below in `second-axis.scss` file :)

darokrk (approved):
I give approval 👍🏽  but look at my mentions and try to fix it before moving next :)

PR #2519, opened 2023-06-13
Initial commit for stopwatch

tania-kuzmenko:
https://tania-kuzmenko.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/blocks/page.scss
+++ b/src/styles/blocks/page.scss
@@ -1,0 +1,8 @@
+.page {
+  box-sizing: border-box;
+
+  &__body {
+    margin: 0;
+    padding: 0;
+  }
+}
--- a/src/styles/blocks/stopwatch.scss
+++ b/src/styles/blocks/stopwatch.scss
@@ -1,0 +1,129 @@
+.stopwatch {
+  position: relative;
+  width: 300px;
+  height: 300px;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  &__center {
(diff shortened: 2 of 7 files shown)


src/styles/blocks/page.scss
```diff
@@ -0,0 +1,8 @@
+.page {
+  box-sizing: border-box;
+
+  &__body {
+    margin: 0;
+    padding: 0;
+  }
+}
```
etojeDenys, line 8: why do you need this classes, if we can just use `html` and `body` selectors

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,129 @@
...
+}
+
+@keyframes minute {
+  0% {
+    transform: rotate(0);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
etojeDenys, line 129: why do we need two the same keyframes

etojeDenys (changes requested):

src/styles/blocks/page.scss
```diff
@@ -0,0 +1,8 @@
+.html {
+  box-sizing: border-box;
+}
+
+.body {
+  margin: 0;
+  padding: 0;
+}
```
VitaliyBondarenko1982, line 8: in these cases better style by tag `body`, `html`

VitaliyBondarenko1982 (approved):
Looks good

PR #2526, opened 2023-06-15
stop watch

yevsss:
[DEMO LINK](https://yevsss.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/animation.scss
+++ b/src/styles/animation.scss
@@ -1,0 +1,27 @@
+@keyframes rotate-hour-arrow {
+  0% {
+    transform: rotate(0deg);
+  }
+
+  50% {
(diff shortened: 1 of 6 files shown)


src/styles/animation.scss
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    50% {
+      transform: rotate(180deg);
+    }
+
+    100% {
+      transform: rotate(360deg);
+    }
+  }
```
etojeDenys, line 27: why do we need two the same keyframes

etojeDenys (changes requested):
small nit

sTorba24 (approved):
GJ 🚀

PR #2530, opened 2023-06-17
add task solution

SerhiiVizir:
- [DEMO LINK](https://SerhiiVizir.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/_clock.scss
+++ b/src/styles/_clock.scss
@@ -1,0 +1,62 @@
+.clock {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+
+  border: 2px dashed black;
+  border-radius: 50%;
+  position: relative;
+
+  width: 80vmin;
+  height: 80vmin;
+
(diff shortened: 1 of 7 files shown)


src/styles/_clock.scss
```diff
@@ -0,0 +1,62 @@
...
+
+@keyframes rotate-second {
+  0% {
+    transform: rotate(0deg);
+  }
+
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
VitaliyBondarenko1982, line 62: why do you need two same keyframes?

VitaliyBondarenko1982 (changes requested):

src/styles/_clock.scss
SerhiiVizir: because there are two arrows

volodymyr-soltys97 (approved):
Well done 👍

PR #2531, opened 2023-06-18
stopwatch has been done

MyNovikov:
[DEMO LINK](https://MyNovikov.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,63 @@
+* {
+  margin: 0;
+}
+
+.circle {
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,63 @@
+* {
+  margin: 0;
```
witflash, line 2: Don't use any rules inside * selector, but box-sizing only.

witflash (approved):
Good job! 👍

PR #2533, opened 2023-06-19
Develop

TenetaJ:
https://TenetaJ.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,8 +1,4 @@
-// $edge_page: calc((100vh - 650px) / 2);
 $height_watch: 80vmin;
-// $width_watch: 80vmin;
-// $center_hand: calc(100vh - (650px / 2 + #{$edge_page}));
-// $center_vertical: calc(50% - (#{$height_watch} / 2));
 
 body {
   display: flex;
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,64 @@
+// $edge_page: calc((100vh - 650px) / 2);
+$height_watch: 80vmin;
+// $width_watch: 80vmin;
+// $center_hand: calc(100vh - (650px / 2 + #{$edge_page}));
+// $center_vertical: calc(50% - (#{$height_watch} / 2));
+
+body {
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,60 @@
...
+
+body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+  margin: auto;
+}
+
+.watch {
```
polosanya, line 11: Create a separate file for watch styles
image

src/styles/main.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  &__second-hand {
+    position: absolute;
+    background-color: green;
+    width: 9px;
+    height: calc(#{$height_watch} / 2);
+    top: 50%;
+    left: 50%;
+    transform-origin: bottom center;
+    animation: rotate 5s linear infinite;
```
polosanya, line 29: It would be great to create variables for animation duration and colors.
image

src/styles/main.scss
```diff
@@ -0,0 +1,60 @@
...
+  margin: auto;
+}
+
+.watch {
+  position: relative;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: $height_watch;
+  height: $height_watch;
```
polosanya, line 17: I guess variable should be named 'watch_size' as it's not only height but width too.
By the way, you could use a mixin for setting same width and height, as you set the identical width and height in watch__round too
https://sass-lang.com/documentation/at-rules/mixin

polosanya (changes requested):
Almost done!

src/styles/main.scss
TenetaJ: done

src/styles/main.scss
TenetaJ: done

src/styles/main.scss
TenetaJ: done

Anastasiia-Svintsova (approved):

PR #2534, opened 2023-06-19
Add task solution

ZakZh:
- [DEMO LINK](https://ZakZh.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/components/watch/watch-animations.scss
+++ b/src/styles/components/watch/watch-animations.scss
@@ -1,0 +1,5 @@
+@keyframes arrow-rotate-anim {
+  to {
+    transform: rotate($watch-steps-size * 1deg * 60);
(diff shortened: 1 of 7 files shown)


src/styles/components/watch/watch-animations.scss
```diff
@@ -0,0 +1,5 @@
+@keyframes arrow-rotate-anim {
+  to {
+    transform: rotate($watch-steps-size * 1deg * 60);
```
etojeDenys, line 3: The rotation duration for the second and minute hands should be 60s and 5s respectively, but the animation is not configured for these durations.

etojeDenys (changes requested):

etojeDenys (approved):
LGTM

PR #2536, opened 2023-06-20
clock

Ishkina:
[DEMO LINK](https://ishkina.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/block/clock.scss
+++ b/src/styles/block/clock.scss
@@ -1,0 +1,53 @@
+html,
+body {
+  height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.clock {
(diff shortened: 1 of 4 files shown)


src/styles/block/clock.scss
```diff
@@ -0,0 +1,53 @@
...
+  align-items: center;
+}
+
+.clock {
+  position: relative;
+  width: 400px;
+  height: 400px;
+  border-radius: 50%;
+  background-color: grey;
+  border: 1px solid black;
```
OleziO, line 15: Use HEX instead of colors name

OleziO (changes requested):

sTorba24 (approved):
GJ 🚀

PR #2566, opened 2023-07-04
Stopwatch. Add task solution

khviktoriia:
- [DEMO LINK](https://khviktoriia.github.io/layout_stop-watch/)


volodymyr-soltys97 (changes requested):
Nice work 👍 
I think it will be better if you center the clock on the page horizontally
image

OleziO (approved):

PR #2572, opened 2023-07-07
add task solution

patryk-babon:
- [DEMO LINK](https://patryk-babon.github.io/layout_stop-watch/)


src/styles/utils/_mixins.scss
choeqq, line 1: I would create separate animation for each hand - seconds hand is looking pretty smooth, when the other 2 are just jumping over few pixels :/

choeqq (changes requested):
Almost there 😄

src/styles/utils/_mixins.scss
patryk-babon: This is a deliberate effect. The hour hand jumps every hour and the minute hand jumps every minute. This is the effect I wanted and I like it.

The animation can be easily customized as you can see in the example below:

```css
@include rotate($seconds-rotation-time, linear);

@include rotate($minutes-rotation-time, steps(60));
```

choeqq (approved):
I don't have any other adjustments, nicely done! 🚀

PR #2579, opened 2023-07-10
layout stop-watch

lisabogdantseva:
- [DEMO LINK](https://lisabogdantseva.github.io/layout_stop-watch/)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1,6 +1,6 @@
 # Stopwatch task
 Replace `<your_account>` with your Github username and copy the links to Pull Request description:
-- [DEMO LINK](https://<your_account>.github.io/layout_stop-watch/)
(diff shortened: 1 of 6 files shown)


lisabogdantseva:
Hi! I've had some problems with npm run lint in this task, the same as in the layout_calendar task. I tried to fix it with npm install --save-dev stylelint-config-recommended-scss, but had no luck with it, so instead I committed with git ci -nm "add stop-watch solution" (I've seen that advice for layout_calendar)

anastasiiavorobiova (changes requested):
Great work! Clock stops, but it should move infinitely

lisabogdantseva:
Thank you for the comments and advice!

mykhalenych (approved):

PR #2588, opened 2023-07-12
add task solution

mykyta01:
- [DEMO LINK](https://mykyta01.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,17 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
         content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="clock">
+      <div class="clock__minute"></div>
+      <div class="clock__second"></div>
+    </div>
+  </body>
 </html>
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,44 @@
+.clock {
+  position: relative;
+  @include square($clockSize);
+  border: 3px dotted black;
+  border-radius: 50%;
+  justify-content: center;
+  align-items: center;
+
+  &::after {
+    content: "";
+    position: absolute;
+    top: 50%;
+    left: 50%;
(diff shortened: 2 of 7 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,44 @@
...
+  &__second {
+    position: absolute;
+    margin: 0 auto;
+    height: 50%;
+    top: 0%;
+    left: 49%;
+    width: 2%;
+    background-color: green;
+    animation: spin 5s linear;
+    transform-origin: bottom center;
```
Esceype, line 42: Find properties that are repeated here and move them to a common class

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,44 @@
...
+  border: 3px dotted black;
+  border-radius: 50%;
+  justify-content: center;
+  align-items: center;
+
+  &::after {
+    content: "";
+    position: absolute;
+    top: 50%;
+    left: 50%;
```
Esceype, line 13: If you use a number / color several times, it is better to move it to a variable. This also applies to all subsequent similar cases

Esceype (changes requested):
GJ!!! But you could also choose any design for this task and add your own specific features.
Also, the stopwatch should be at the center of the page.

src/index.html
```diff
@@ -1,14 +1,17 @@
...
-</body>
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <title>Stop watch</title>
+    <link rel="stylesheet" href="./styles/main.scss">
+  </head>
+  <body>
+    <div class="clock">
+      <div class="clock__minute"></div>
+      <div class="clock__second"></div>
+    </div>
```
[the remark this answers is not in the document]
mykyta01: Ok, what is better option?

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,45 @@
+.clock {
+  position: relative;
+  @include square($clockSize);
+  border: 3px solid black;
+  margin: 5% auto;
+  justify-content: center;
+  align-items: center;
+  background-color: #efd6d6;
```
denys-danyliuk, line 8: 8'th point in checklist

denys-danyliuk (changes requested):

src/styles/blocks/clock.scss
mykyta01: I understand that, but in that case idk what in that properties  i should move into another class

Esceype (changes requested):
@mykyta01 Read point number 8 in the checklist.  There is written about how to solve the problem that you say you don't know how to solve and about which @denys-danyliuk wrote to you.
Create a container(html) for your clock and move all external styles (positioning or margins) to styles inside the class you define for the container

mykyta01:
sry was inattentive

src/styles/main.scss
```diff
@@ -0,0 +1,9 @@
+@import "./utils/variables.scss";
+ @import "./utils/mixins.scss";
+ @import "./utils/animations.scss";
+
+ @import "./blocks/clock.scss";
+
+ * {
+   box-sizing: border-box;
+ }
```
Andriy-Shymkiv, line 9: move this styles to a separate file

Andriy-Shymkiv (approved):
good job!

PR #2591, opened 2023-07-12
Develop

yuran007:
[DEMO LINK](https://yuran007.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,0 +1,74 @@
+body {
+  margin: 0;
+  height: 100vh;
+
+  display: grid;
+  place-content: center;
+}
+
+.watch {
+  --size: 80vmin;
+  position: relative;
+
+  height: var(--size);
+  width: var(--size);
+
+  border-radius: 50%;
+  border: calc(var(--size) / 60) dotted black;
+
+  &::before,
+  &::after {
+    content: "";
+    position: absolute;
+    top: 50%;
+    left: 50%;
+
+    transform: translate(-50%, -50%) rotate(-90deg);
+
+    display: block;
+    transform-origin: left;
+  }
+
+  &::before {
+    height: 5px;
+    width: calc(var(--size) / 2);
+
+    background-color: blue;
+
+    animation: 5s move steps(60);
+  }
+
+  &::after {
+    height: 10px;
+    width: calc(var(--size) / 4);
+
+    background-color: green;
+
+    animation: 60s move steps(60);
+  }
+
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -0,0 +1,74 @@
...
+    transform-origin: left;
+  }
+
+  &::before {
+    height: 5px;
+    width: calc(var(--size) / 2);
+
+    background-color: blue;
+
+    animation: 5s forwards move steps(60);
```
Viktor-Kost, line 38: animations should be infinite

src/styles/main.scss
```diff
@@ -0,0 +1,74 @@
...
+    position: absolute;
+    top: 50%;
+    left: 50%;
+
+    transform: translate(-50%, -50%);
+
+    width: 20px;
+    height: 20px;
+
+    border-radius: 50%;
```
Viktor-Kost, line 60: since you use circle twice you can just create a mixin for watch and watch__center

```
@mixin circle($width) {
  width: $width;
  height: $width;
  border-radius: $width/2;
}

.watch {
  @include circle(var(--size));
}
```

Viktor-Kost (changes requested):
Great job!
Some minor fixes needed

src/styles/watch.scss
```diff
@@ -0,0 +1,52 @@
+.watch {
+  --size: 80vmin;
```
OleziO, line 2: Move to :root selector.

src/styles/watch.scss
```diff
@@ -0,0 +1,52 @@
...
+
+    display: block;
+    transform-origin: left;
+  }
+
+  &::before {
+    height: 5px;
+    width: calc(var(--size) / 2);
+
+    background-color: blue;
```
OleziO, line 26: Use HEX instead of color names.

OleziO (changes requested):

src/styles/watch.scss
yuran007: Done

src/styles/watch.scss
yuran007: Done

src/styles/main.scss
yuran007: Done

src/styles/main.scss
yuran007: Done

demetrxx (approved):
Well done 👍 Go ahead!

PR #2594, opened 2023-07-14
add clock with scss animation 

FrontKid:
- [DEMO LINK](https://FrontKid.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/1-helpers/_variables.scss
+++ b/src/styles/1-helpers/_variables.scss
@@ -1,0 +1,9 @@
+// colors
+$white: #fff;
+$blue: #00acdc;
+$black: #000;
+$grey: #eee;
+
+// general
+$fz: 30px;
+$lh: 100%;
--- a/src/styles/2-components/_watch.scss
+++ b/src/styles/2-components/_watch.scss
@@ -1,0 +1,36 @@
+.watch {
+  border: 4vmin dotted black;
+  border-radius: 50%;
+  min-width: 80vmin;
+  min-height: 80vmin;
+
+  position: relative;
+}
+
+.watch__second {
+  width: 4vmin;
+  height: 40vmin;
+  @include arrowSettings (green, second, 5s);
+}
+
+.watch__minute {
+  width: 8vmin;
+  height: 20vmin;
+  @include arrowSettings (blue, minute, 60s);
+}
+
+.watch__center {
+  background-color: yellow;
+  width: 13vmin;
+  height: 13vmin;
+  border-radius: 50%;
+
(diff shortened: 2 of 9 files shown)


src/styles/1-helpers/_variables.scss
```diff
@@ -0,0 +1,9 @@
+// colors
+$white: #fff;
+$blue: #00acdc;
+$black: #000;
+$grey: #eee;
+
+// general
+$fz: 30px;
+$lh: 100%;
```
alexander-ignatow, line 9: Pick readable variable names

alexander-ignatow (changes requested):
Great solution!

Just rename these variables :)

src/styles/1-helpers/_variables.scss
```diff
@@ -0,0 +1,9 @@
+// colors
+$white: #fff;
+$blue: #00acdc;
+$black: #000;
+$grey: #eee;
```
roman-mirzoian, line 5: It is better to name the variables not according to the content, but according to the meaning, because the color may change during the development process and it will have to be corrected in the entire file.

roman-mirzoian (changes requested):

src/styles/1-helpers/_variables.scss
FrontKid: done

src/styles/2-components/_watch.scss
```diff
@@ -0,0 +1,36 @@
...
+  width: 13vmin;
+  height: 13vmin;
+  border-radius: 50%;
+
+  position: absolute;
+  left: 50%;
+  top: 50%;
+
+  transform: translate(-50%,-50%);
+}
```
roman-mirzoian, line 33: Use nesting where possible.
https://sass-lang.com/guide/#nesting

src/styles/1-helpers/_variables.scss
```diff
@@ -0,0 +1,9 @@
+// colors
+$point-color: #000;
+$sec-arrow-color: green;
+$center-arrow-color: yellow;
+$min-arrow-color: blue;
+
+// general
+$mainFontSize: 30px;
+$mainLineHeight: 100%;
```
roman-mirzoian, line 9: When naming variables, you should follow the same style code for all variables.

roman-mirzoian (changes requested):

src/styles/1-helpers/_variables.scss
FrontKid: done

PR #2596, opened 2023-07-14
add task solution

MKLNKs:
- [DEMO LINK](https://MKLNKs.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,56 @@
+.watch {
+  @include contentFlexCenterCenter;
+
+  width: $watch-size;
+  height: $watch-size;
+
+  background-color: #33867e;
+  box-shadow: 1px 5px 7px #1c1c1c;
+  border-radius: 50%;
+
+  &__center {
+    z-index: 1;
+
+    width: 0.03 * $watch-size;
+    height: 0.03 * $watch-size;
+
+    box-shadow: $elements-shadow;
+    background-color: $center-bg-color;
+    border: 2px solid $minute-color;
+    border-radius: 50%;
+  }
+
+  &__minutes {
+    @include placeOfHands;
+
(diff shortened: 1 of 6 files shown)


src/styles/utils/variables.scss
```diff
@@ -0,0 +1,4 @@
+$watch-size: 80vmin;
```
volodymyr-soltys97, line 1: The variables defined are correct as per the task description, but it's missing a variable for the second hand color.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,56 @@
...
+    @include placeOfHands;
+
+    width: 0.01 * $watch-size;
+    height: 0.4 * $watch-size;
+
+    box-shadow: $elements-shadow;
+    background-color: $minute-color;
+    border-radius: 10px;
+
+    @include rotateSettings(3660s);
```
volodymyr-soltys97, line 33: The rotation settings for the minute hand seem to be incorrect. As per the task, the minute hand should rotate every 60 seconds, not every 3660 seconds.

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,56 @@
...
+  &__seconds {
+    @include placeOfHands;
+
+    width: 0.003 * $watch-size;
+    height: 0.49 * $watch-size;
+
+    box-shadow: $elements-shadow;
+    background-color: $center-bg-color;
+
+    @include rotateSettings(60s);
```
volodymyr-soltys97, line 45: The rotation settings for the second hand seem to be correct in terms of time (60 seconds), but you need to ensure that it starts from the top. You might need to rotate it -90 degrees initially to start from a vertical position.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

PR #2599, opened 2023-07-15
add task solution

ndovgyi:
[DEMO LINK] https://ndovgyi.github.io/layout_stop-watch/


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,23 @@
+@import "../veriables.scss";
+
+.clock {
+  box-sizing: border-box;
```
OleziO, line 4: Move to the * selector and remove it from all other places.

OleziO (changes requested):
Align arrows to the center
image

src/styles/blocks/clock.scss
OleziO: Not resolved

OleziO (changes requested):

src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,26 @@
+@import "../veriables.scss";
+
+* {
```
mykhalenych, line 3: better will be inside main.scss file

mykhalenych (approved):

PR #2601, opened 2023-07-15
add task solution

MKLNKs:
- [DEMO LINK](https://MKLNKs

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,20 @@
-<!doctype html>
+<!DOCTYPE html>
 <html lang="en">
   <head>
-    <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <meta charset="UTF-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Stop watch</title>
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="stylesheet" href="./styles/main.scss" />
   </head>
-  <body>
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,14 +1,20 @@
...
-    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <meta charset="UTF-8" />
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Stop watch</title>
-    <link rel="stylesheet" href="./styles/main.scss">
+    <link rel="stylesheet" href="./styles/main.scss" />
```
TarasHoliuk, line 11:
```suggestion
    <link rel="stylesheet" href="./styles/main.scss">
```
fix everywhere in the head:
image

src/index.html
```diff
@@ -1,14 +1,20 @@
...
   </head>
-  <body>
-    <h1>Stop watch</h1>
+  <body class="page">
+    <section class="page__watch watch">
+      <div class="watch__center"></div>
```
TarasHoliuk, line 15:
```suggestion
      <div class="watch__center" />
```
but here you may use self-close elements

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,15 @@
+@mixin contentFlexCenterCenter() {
```
TarasHoliuk, line 1:
```suggestion
@mixin contentFlexCenter {
```
you may remove brackets if mixin doesn't accept any argument

TarasHoliuk (changes requested):

src/index.html
MKLNKs: I have set up Linter and Prettier this way purposely, that's why I can't fix this mistake. 
I have done my previous tasks with this problem..

src/index.html
MKLNKs: Also.. Linter doesn't miss.. But, thanks!)

src/styles/utils/mixins.scss
MKLNKs: Thank you for recommendations!

PR #2614, opened 2023-07-20
add task solution

voievodik:
[DEMO LINK](https://voievodik.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,13 +8,11 @@
     <title>Stop watch</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <div class="wrapper">
-      <div class="watch">
-        <div class="watch__minute-arrow"></div>
-        <div class="watch__hour-arrow"></div>
-        <div class="watch__circle"></div>
-      </div>
+  <body class="wrapper">
+    <div class="watch">
+      <div class="watch__minute-arrow"></div>
+      <div class="watch__hour-arrow"></div>
+      <div class="watch__circle"></div>
     </div>
   </body>
 </html>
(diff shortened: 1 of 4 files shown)


src/index.html
```diff
@@ -9,6 +9,12 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Stop watch</h1>
+    <div class="wrapper">
```
roman-mirzoian, line 12: You don't need to wrap the content in an extra `div` block, you can set a `wrapper` class for the `body` and customize it.

src/styles/main.scss
```diff
@@ -1 +1,61 @@
 // put code here
+@import "./utils/mixins.scss";
+
+$size-watch: 80vmin;
+$percent-value: 50%;
+
+body {
```
roman-mirzoian, line 7: Try to specify styling through classes where possible, rather than working with tags directly. For example, you now have a `wrapper` class in HTML, but you don't style it at all.

src/styles/main.scss
```diff
@@ -1 +1,61 @@
 // put code here
+@import "./utils/mixins.scss";
+
+$size-watch: 80vmin;
+$percent-value: 50%;
```
roman-mirzoian, line 5: Try to name variables not by content, but by meaning, because the content can change (for example, percentages will change on pixels) during development and you will have to fix it in the entire code.

src/styles/main.scss
```diff
@@ -1 +1,61 @@
...
+    width: $size-watch / 15;
+    height: $size-watch / 15;
+    transform: translate(-$percent-value, -$percent-value);
+    border-radius: $percent-value;
+    background-color: #d0629c;
+    @include position($percent-value);
+  }
+}
+
+@keyframes round {
```
roman-mirzoian, line 54: The animation of the clock is better called `tick` or `arrowRotation`

roman-mirzoian (approved):
Well done 🔥

PR #2624, opened 2023-07-24
add task solution

viktoria-roik:
- [DEMO LINK](https://viktoria-roik.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -1,14 +1,18 @@
 <!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport"
-        content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
-  <meta http-equiv="X-UA-Compatible" content="ie=edge">
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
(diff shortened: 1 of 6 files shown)


src/index.html
```diff
@@ -1,14 +1,18 @@
...
-  <title>Stop watch</title>
-  <link rel="stylesheet" href="./styles/main.scss">
-</head>
-<body>
-  <h1>Stop watch</h1>
-</body>
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport"
+          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
```
roman-mirzoian, line 6: image

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,4 @@
+$size-of-clock: 80vmin;
```
roman-mirzoian, line 1:
```suggestion
$clock-size: 80vmin;
```

src/styles/utils/_variables.scss
```diff
@@ -0,0 +1,4 @@
...
+$size-of-circle: 20px;
```
roman-mirzoian, line 2:
```suggestion
$circle-size: 20px;
```

roman-mirzoian (changes requested):
Also, try to align the arrows more precisely so that they are "pinned" exactly in the center:
image

src/styles/utils/_mixins.scss
```diff
@@ -0,0 +1,8 @@
+@mixin generalProperties {
```
Anna-Tkachenko, line 1: Will be better to use more clear name for the mixin
I can't get from the name for what element and for what purpose I can add it

Anna-Tkachenko (approved):

PR #2627, opened 2023-07-24
add task solution

davinnchii:
[DEMO](https://davinnchii.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -1 +1,70 @@
 // put code here
+
```
roman-mirzoian, line 2: Do not leave extra empty lines.

src/styles/main.scss
```diff
@@ -1 +1,70 @@
 // put code here
```
roman-mirzoian, line 1: Do not leave unnecessary comments.

src/styles/main.scss
```diff
@@ -1 +1,70 @@
...
+    position: absolute;
+    top: 50%;
+    left: 50%;
+    transform: translate(-50%, -100%) rotate(0deg);
+    transform-origin: bottom;
+
+    animation-name: arrow-spin;
+    animation-duration: 60s;
+    animation-timing-function: linear;
+    animation-iteration-count: infinite;
```
roman-mirzoian, line 56: You now have the parameters for `transform` and `animation` repeated twice, only with a difference in `duration`. Try to use the features of SCSS to avoid repeating the code unnecessarily.

roman-mirzoian (changes requested):

PR #2641, opened 2023-07-29
add task solution

AndriiZhovtko:
- [DEMO LINK](https://AndriiZhovtko.github.io/layout_stop-watch/)


src/styles/main.scss
```diff
@@ -1 +1,72 @@
-// put code here
+body {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin: 0 auto;
+}
+
+.clock {
```
ivbutsykin, line 8: Move BEM component styles in separate `scss` file

src/styles/main.scss
```diff
@@ -1 +1,72 @@
-// put code here
+body {
```
ivbutsykin, line 1: Better to create separate `div` with class `.wrapper` or sth else because doesn't good set such classes to `body`

src/styles/main.scss
```diff
@@ -1 +1,72 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin: 0 auto;
+}
+
+.clock {
+  position: relative;
+  display: flex;
+  width: 80vmin;
```
ivbutsykin, line 11: Create file with constants and add sizes there

src/styles/main.scss
```diff
@@ -1 +1,72 @@
...
+  height: 80vmin;
+  border: 3px solid #000;
```
ivbutsykin, line 13: Same for colors (move to constants)

ivbutsykin (changes requested):
Overall, good work, but need add some fixes

src/styles/clock.scss
```diff
@@ -0,0 +1,65 @@
...
+}
+
+@keyframes second {
+  from {
+    rotate: 0deg;
+  }
+  to {
+    rotate: 360deg;
+  }
+}
```
etojeDenys, line 65: why do you need two the same keyframes?

etojeDenys (changes requested):

SanyaBratashchuk (approved):

PR #2643, opened 2023-07-30
add task solution

OlhaMomot:
[DEMO LINK](https://OlhaMomot.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,65 @@
-// put code here
+$clock-size: 80vmin;
+$middle-size: 8vmin;
+$arrow-size: 2vmin;
+
+.container {
+  width: fit-content;
+  margin: auto;
+}
+
+.clock {
+  position: relative;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,65 @@
...
+  height: $clock-size;
+  width: $clock-size;
+  border: 20px double rebeccapurple;
+  background-color: white;
+  border-radius: 50%;
+
+  &__arrow {
+    position: absolute;
+    left: $clock-size / 2 - $arrow-size / 2;
+    width: $arrow-size;
```
Anna-Tkachenko, line 21: Will be better to move calculations to separate variables

Anna-Tkachenko (approved):
Some comments:
- will be great to play with animation to make it work smoothly
- you didn't create any variables, that's not good approach. Will be good to move all calculations and main sizes to the separate variables

PR #2647, opened 2023-07-30
add task solution

desevoker:
- [DEMO LINK](https://desevoker.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/utils/variables.scss
+++ b/src/styles/utils/variables.scss
@@ -1,0 +1,26 @@
+$size: 80vmin;
+$border-size: $size / 80;
+$half-size: $size / 2;
+
+$knob-size: $half-size / 12;
+$knob-offset: $half-size - ($knob-size / 2);
+
+$min-arm-width: $size / 32;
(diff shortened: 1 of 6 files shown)


src/styles/utils/variables.scss
```diff
@@ -0,0 +1,26 @@
...
+
+$knob-size: $half-size / 12;
+$knob-offset: $half-size - ($knob-size / 2);
+
+$min-arm-width: $half-size / 16;
+$min-arm-height: $half-size / 1.1;
+$min-arm-offset: $min-arm-height / 8;
+$min-arm-top: $half-size - $min-arm-height + $min-arm-offset;
+$min-arm-left: $half-size - ($min-arm-width / 2);
+$min-arm-origin: center ($min-arm-height - $min-arm-offset);
```
Anna-Tkachenko, line 13: awesome job with variables calculation! With such approach we can make sure that everything will work ok if we want to change the size for the watch

Anna-Tkachenko (approved):

PR #2649, opened 2023-07-31
task solution

MarekSax:
[DEMO](https://mareksax.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/container.scss
+++ b/src/styles/container.scss
@@ -1,0 +1,26 @@
+.container {
+  height: 100vh;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: relative;
+
+  // &::before {
+  //   content: '';
+  //   height: 1px;
+  //   width: 100%;
+  //   background-color: black;
+  //   z-index: 3;
+
+  // }
+
+  // &::after {
+  //   content: '';
+  //   position: fixed;
+  //   left: calc(50% - 0.5px);
+  //   height: 100%;
+  //   width: 1px;
+  //   background-color: black;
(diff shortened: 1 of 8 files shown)


src/styles/container.scss
```diff
@@ -0,0 +1,26 @@
+.container {
+  height: 100vh;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  position: relative;
+
+  // &::before {
```
darokrk, line 8: You have commented out code starting from this line. If this code is not necessary, then it would be best to remove it to avoid confusion and keep the codebase clean.

src/styles/main.scss
```diff
@@ -1 +1,6 @@
-// put code here
+@import "../utils/variables";
+@import "../utils/reset";
+@import "../utils/keyframes";
+
+@import "./container.scss";
+@import "./stop-watch.scss";
```
darokrk, line 6: Once again, avoid using relative paths in import statements. Use absolute paths for a clear project structure and easier refactoring.

src/styles/main.scss
```diff
@@ -1 +1,6 @@
-// put code here
+@import "../utils/variables";
+@import "../utils/reset";
```
darokrk, line 2: the reset import should be used in 1st line because rest of files can depend on that, so please move this declaration up

darokrk (changes requested):
Nice UI 🔥 ! well done 👏🏽 two minor fixes and we are good to go 💪🏽 !

darokrk (approved):
Well done 🏆 !

PR #2653, opened 2023-08-01
very basic stop watch 1

steveforde:
- [DEMO LINK](https://steveforde.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,58 @@
-// put code here
+* {
+  margin: 0;
+}
+
+body {
+  padding: 0;
+  box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,58 @@
-// put code here
+* {
+  margin: 0;
```
witflash, line 2: Don't use any rules inside `*` selector instead of `box-sizing`. Place `margin` everywhere you need but not do it globally.

witflash (changes requested):
Good job!
Let's improve a little bit. Pay attention to my comments, please.

PR #2655, opened 2023-08-02
ready to check

robsza1122:
Before the code review: 
 
  Firstly, I couldn't properly set layout on 80 vmin size. I saw other students did watch-size on pixels and it was OK so I did it as well. Secondly, I couldn't set in border's watch 60 dots. Is it possible to do?

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -6,9 +6,31 @@
           content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Stop watch</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com"
+    crossorigin="anonymous">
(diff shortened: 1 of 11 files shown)


src/styles/center.scss
```diff
@@ -0,0 +1,9 @@
+.watch {
```
darokrk, line 1: just simplify this selector

```
.watch__center {
  ......
}
```

src/index.html
```diff
@@ -6,9 +6,31 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Rye&display=swap" rel="stylesheet">
+
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Stop watch</h1>
+    <section class="watch">
+      <p class="watch__center"></p>
+      <p class="watch__number_1">1</p>
+      <p class="watch__number_2">2</p>
```
darokrk, line 20: those indicators can be rather `<span>` elements bcs they don't have any meaningful content

darokrk (changes requested):
UI looks good 👍🏽  but code need few more improvements 🛠️

Radoslaw-Czerniawski (approved):
Approved, watch looks good :D

PR #2663, opened 2023-08-04
done

ikispower:
[DEMO LINK](https://ikispower.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,67 @@
-// put code here
+body {
+  position: relative;
+  min-height: 100vh;
+  margin: 0;
+  padding: 0;
+}
+
+.stopwatch {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  width: 80vmin;
+  height: 80vmin;
+  border: 10px dotted #000;
+  border-radius: 50%;
+
+  &__center {
+    position: absolute;
+    left: 50%;
+    top: 50%;
+    transform: translate(-50%, -50%);
+    background-color: black;
+    height: 30px;
+    width: 30px;
+    border-radius: 50%;
+  }
+}
+
+.hand {
+  position: absolute;
+  left: 50%;
+  top: 50%;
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1 +1,67 @@
-// put code here
+body {
+  position: relative;
+  min-height: 100vh;
+  margin: 0;
+  padding: 0;
+}
+
+.stopwatch {
```
TarasHoliuk, line 8: let's create a separate file for this BEM-block

src/styles/main.scss
```diff
@@ -1 +1,67 @@
...
+    top: 50%;
+    transform: translate(-50%, -50%);
+    background-color: black;
+    height: 30px;
+    width: 30px;
+    border-radius: 50%;
+  }
+}
+
+.hand {
```
TarasHoliuk, line 30: if it's a separate block it should be moved to a separate file
or make it an element of stopwatch block (use modificators if you need)

TarasHoliuk (changes requested):

volodymyr-soltys97 (changes requested):
The arrows are not placed exactly in the center, it needs to be fixed
image

src/styles/main.scss
```diff
@@ -1 +1,26 @@
...
+}
+
+@keyframes sec {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
```
VitaliyBondarenko1982, line 26: no need to make two same keyframes.
you can use one keyframes for both arrows

VitaliyBondarenko1982 (approved):
Good job.

src/styles/main.scss
anastasiiavorobiova: Consider resolving this comment

anastasiiavorobiova (changes requested):
Great work! Almost done!

volodymyr-soltys97 (approved):
Well done 🔥

PR #2670, opened 2023-08-08
add task solution

SviatlanaRamanauskas:
- [DEMO LINK](https://SviatlanaRamanauskas.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,10 @@
-// put code here
+body {
+  margin: 0;
+  height: 100vh;
+  display: flex;
+}
+
+@import "./utils/variables";
+@import "utils/functions";
+@import "utils/mixins";
+@import "blocks/clock";
--- a/src/styles/utils/variables.scss
+++ b/src/styles/utils/variables.scss
@@ -1,0 +1,6 @@
+$clockDiameter: 500px;
+$clockRadius: 250px;
+$halfClockRadius: $clockRadius / 2;
+$hourWidth: 12px;
+$minuteWidth: 8px;
+$centerCircleWidth: 50px;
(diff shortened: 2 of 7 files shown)


src/styles/main.scss
```diff
@@ -1 +1,10 @@
-// put code here
+body {
+  margin: 0;
+  height: 100vh;
+  display: flex;
+}
+
+@import "./utils/variables";
+@import "utils/functions";
```
volodymyr-soltys97, line 8: You are missing the dot before the path, causing the file to not be found. It should be './utils/functions'.

src/styles/main.scss
```diff
@@ -1 +1,10 @@
-// put code here
+body {
+  margin: 0;
+  height: 100vh;
+  display: flex;
+}
+
+@import "./utils/variables";
```
volodymyr-soltys97, line 7: You are missing the dot before the path, causing the file to not be found. It should be './utils/variables'.

src/styles/main.scss
```diff
@@ -1 +1,10 @@
...
+@import "utils/functions";
+@import "utils/mixins";
+@import "blocks/clock";
```
volodymyr-soltys97, line 10: You are missing the dot before the path, causing the file to not be found. It should be './blocks/clock'.

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,6 @@
+$clockDiameter: 500px;
```
volodymyr-soltys97, line 1: The task description specifies that the size of the stopwatch should be 80vmin, but you have set a fixed pixel value. This might not make the stopwatch responsive. Consider using 'vmin' unit instead of 'px' for sizing.

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Outstanding work!

PR #2686, opened 2023-08-13
add layout stop-watch

Dizelkos:


denys-danyliuk (changes requested):
Not all requirements are met 

Знімок екрана 2023-08-14 о 10 54 38

Before making a Pull Request, you need to check the task description and make sure that all requirements are met. After that, the mentor will be able to review the homework. If you have questions please ask them in fe_chat

PR #2688, opened 2023-08-14
add solution task

Butovych:
[DEMO LINK](https://Butovych.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/utils/variables.scss
+++ b/src/styles/utils/variables.scss
@@ -1,0 +1,28 @@
+$clock-height: 80vmin;
+$clock-width: 80vmin;
+
+$clock__centre-height: 5vmin;
+$clock__centre-width: 5vmin;
+
+$clock__minute-hand-height: 30vmin;
+$clock__minute-hand-width: 3vmin;
+
+$clock__second-hand-height: $clock-height / 2;
+$clock__second-hand-width: 1vmin;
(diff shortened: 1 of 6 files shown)


src/styles/utils/variables.scss
```diff
@@ -0,0 +1,28 @@
+$clock-height: 80vmin;
```
mykhalenych, line 1: why do you need two variables with the same value?

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,28 @@
...
+$clock-width: 80vmin;
+
+$clock__centre-height: 5vmin;
```
mykhalenych, line 4: the same

mykhalenych (changes requested):
nice work, but fix comments

src/styles/utils/variables.scss
Butovych: Thanks for the answer. I fixed this error. But I was guided by the video "When to use variables" in the topic about sass (1 min, 20 sec)

src/styles/utils/variables.scss
Butovych: ✅

Andriy-Shymkiv (approved):

PR #2694, opened 2023-08-17
add task solution

Xsaites:


src/styles/main.scss
```diff
@@ -1 +1,53 @@
...
+  }
+}
+
+@keyframes Second {
+  100% {
+    transform: translateX(-100%) translateY(-50%) rotate(360deg);
+  }
+}
+
+@keyframes Minute {
```
mykhalenych, line 49: why do you create two similar keyframes?
use one instead

src/styles/main.scss
```diff
@@ -1 +1,53 @@
...
+  animation-duration: 5s;
+  animation-timing-function: linear;
+  animation-iteration-count: infinite;
+
+  &--minute {
+    background-color: blue;
+    width: 40%;
+    height: 3px;
+    animation: Minute;
+    animation-duration: 60s;
```
mykhalenych, line 39: do it infinite as well

mykhalenych (approved):
Nice work, but fix small suggestions

PR #2701, opened 2023-08-18
Develop

leravlasova:


src/styles/main.scss
```diff
@@ -1 +1,73 @@
...
+@keyframes moveMinute {
+  from {
+    transform: translateX(-50%) rotate(0deg);
+  }
+  to {
+    transform: translateX(-50%) rotate(360deg);
+  }
+}
+
+@keyframes moveSecond {
```
mykhalenych, line 66: why do you need two similar @keyframes, rename and use only one

mykhalenych (changes requested):

nazarmatsevych (approved):

PR #2704, opened 2023-08-19
add task solution

Phetryska:
- [DEMO LINK](https://Phetryska.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/stopwatch.scss
+++ b/src/styles/stopwatch.scss
@@ -1,0 +1,60 @@
+body {
+  margin: 0;
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  background-color: $bodyBackground-color;
+}
+
+.watch {
+  display: flex;
+  width: $watch-size;
+  height: $watch-size;
+  box-shadow: 5px 7px 10px -1px rgba(166, 166, 166, 0.25) inset,
+    -6px -1px 10px 2px rgba(255, 255, 255, 0.5) inset,
+    -6px -7px 20px 0 rgba(255, 255, 255, 0.5),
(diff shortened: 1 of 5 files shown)


src/styles/stopwatch.scss
```diff
@@ -0,0 +1,60 @@
...
+  &__center {
+    z-index: 1;
+    width: $watchCenter-size;
+    height: $watchCenter-size;
+    background-color: #272727;
+    border-radius: $radius;
+  }
+
+  &__second-hand {
+    position: absolute;
```
Andriy-Shymkiv, line 32: looks good but I don't see `position: relative;` on parent element. where do you position that from?

src/styles/stopwatch.scss
```diff
@@ -0,0 +1,60 @@
...
+}
+
+.watch {
+  display: flex;
+  width: $watch-size;
+  height: $watch-size;
+  box-shadow: 5px 7px 10px -1px rgba(166, 166, 166, 0.25) inset,
+    -6px -1px 10px 2px rgba(255, 255, 255, 0.5) inset,
+    -6px -7px 20px 0 rgba(255, 255, 255, 0.5),
+    8px 8px 20px 0 rgba(170, 170, 170, 0.25);
```
Andriy-Shymkiv, line 17: you probably can move it to some variable for better readability

Andriy-Shymkiv (changes requested):

src/styles/stopwatch.scss
Phetryska: > looks good but I don't see `position: relative;` on parent element. where do you position that from?

It is not clear why, but it positioned elements relative to the parent element without this setting

Viktor-Kost (approved):

PR #2712, opened 2023-08-21
add layout stop-watch solution

MBohatova:
[DEMO LINK](https://MBohatova.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,59 @@
-// put code here
+$displaySize: 240px;
+$minuteHeight: $displaySize / 2;
+$minuteWidth: 7px;
+$hourHeight: $minuteHeight / 2;
+$hourWidth: 12px;
+$centerSize: 15px;
+
+* {
+  box-sizing: border-box;
+}
+
+.display {
+  position: relative;
+  height: $displaySize;
+  width: $displaySize;
+  margin: auto;
+  border: 5px dotted #000;
+  border-radius: 50%;
+
+  &__minute {
+    position: absolute;
+    bottom: 48%;
+    left: 49.9%;
+    height: $minuteHeight;
+    width: $minuteWidth;
+    background-color: #f09;
+    transition: transform 5s;
+
+    &:hover {
+      transform: rotate(360deg);
+      transform-origin: bottom;
+    }
+  }
+
+  &__hour {
+    position: absolute;
+    bottom: 50%;
+    left: 48.5%;
+    height: $hourHeight;
+    width: $hourWidth;
+    background-color: #0f7;
+    transition: transform 60s;
+
+    &:hover {
+      transform: rotate(360deg);
+      transform-origin: bottom;
+    }
+  }
+
+  &__center {
+    position: absolute;
+    bottom: 46%;
+    left: 48%;
+    height: $centerSize;
+    width: $centerSize;
+    border-radius: 50%;
+    background-color: #f75;
+  }
+}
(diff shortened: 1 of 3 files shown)


witflash (changes requested):
Good design 👍🏻
But you should implement `animation` not `hover` state. So create `@keyframes` and apply this rule as animation for the both hands.

src/styles/main.scss
```diff
@@ -1 +1,76 @@
-// put code here
+$displaySize: 240px;
+$minuteHeight: $displaySize / 2;
+$minuteWidth: 7px;
+$hourHeight: $minuteHeight / 2;
+$hourWidth: 12px;
+$centerSize: 15px;
```
volodymyr-soltys97, line 6: Move it to variables file

src/styles/main.scss
```diff
@@ -1 +1,76 @@
...
+      from {
+        transform: rotate(0deg);
+        transform-origin: bottom;
+      }
+
+      to {
+        transform: rotate(360deg);
+        transform-origin: bottom;
+      }
+    }
```
volodymyr-soltys97, line 63: Now you are duplicating this animation twice, in this case you can get away with one

volodymyr-soltys97 (changes requested):
Good job 👍 
Need to fix the size as on the design
Also, the animation should be endless
image
image

src/styles/main.scss
Viktor-Kost: isn't fixed. 
1. You need to create only one animation and reuse it both to minute hand an hour one.
2. Also you shouldn't use  transform-origin: bottom; inside animation.
3. Move keyframes outside selectors

src/styles/main.scss
```diff
@@ -1 +1,73 @@
...
+  &__hour {
+    position: absolute;
+    bottom: 50%;
+    left: 48.5%;
+    height: $hourHeight;
+    width: $hourWidth;
+    background-color: #0f7;
+    animation-name: move;
+    animation-duration: 8s;
+    animation-iteration-count: infinite;
```
Viktor-Kost, line 48: apply animation `shorthand`

src/styles/main.scss
```diff
@@ -1 +1,73 @@
...
+
+  &__hour {
+    position: absolute;
+    bottom: 50%;
+    left: 48.5%;
+    height: $hourHeight;
+    width: $hourWidth;
+    background-color: #0f7;
+    animation-name: move;
+    animation-duration: 8s;
```
Viktor-Kost, line 47: 60sec

src/styles/main.scss
```diff
@@ -1 +1,73 @@
...
+        transform: rotate(360deg);
+        transform-origin: bottom;
+      }
+    }
+  }
+
+  &__hour {
+    position: absolute;
+    bottom: 50%;
+    left: 48.5%;
```
Viktor-Kost, line 42: try to work with positioning and transforming
Screenshot 2023-12-16 at 15 29 56

Viktor-Kost (changes requested):

MBohatova:
Are you kidding me?.

сб, 16 груд. 2023 р. о 15:30 Viktor-Kost ***@***.***> пише:

volodymyr-soltys97 (changes requested):
Good job 👍 
Still not fixed from the previous review
Visually, it now looks like the arrows are not located in the center of the circle, this needs to be fixed
image

volodymyr-soltys97 (approved):
Great work 🔥

PR #2737, opened 2023-09-04
Develop

Kuper18:
Hello!

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,10 +17,13 @@
           <div class="watch__content">
             <div class="watch__animation watch__animation--hours">
               <div class="watch__circle"></div>
+
               <div class="watch__hours-arrow"></div>
             </div>
+
             <div class="watch__animation watch__animation--seconds">
               <div class="watch__circle"></div>
+
               <div class="watch__seconds-arrow"></div>
             </div>
           </div>
--- a/src/index.html
+++ b/src/index.html
@@ -2,13 +2,30 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8">
-    <meta name="viewport"
-          content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+    <meta
+      name="viewport"
+      content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0"
+    >
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <title>Stop watch</title>
-    <link rel="stylesheet" href="./styles/main.scss">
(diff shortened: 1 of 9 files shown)


src/styles/utils/variables.scss
sviatikkondur, line 1: Unnecessary variables, don't create variable if you use it only one time and if you want to make some sizes dependent on the parent element, just use %

sviatikkondur (changes requested):

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,16 @@
+$size-container-watch: 500px;
```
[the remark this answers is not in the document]
Kuper18: yes, but it was said that there is not strict design

src/styles/utils/variables.scss
Kuper18: I did it because it is easy to manipulate the whole object. In my case, I can't set value in %, it won't be displayed at all

src/index.html
```diff
@@ -2,13 +2,29 @@
...
+          <div class="watch__hours-arrow"></div>
+        </div>
+
+        <div class="watch__animation watch__animation--seconds">
+          <div class="watch__circle"></div>
+
+          <div class="watch__seconds-arrow"></div>
+        </div>
+      </div>
+    </main>
```
StasSokolov, line 28: too complicated code, you can left only one circle in  `watch__animation watch__animation--seconds` element

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,52 @@
...
+
+@keyframes animate-arrow {
+  0% {
+    transform: rotate(0);
+  }
+
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
StasSokolov, line 52: move to the separated file, like `keyframes.scss`

src/styles/utils/variables.scss
```diff
@@ -0,0 +1,16 @@
...
+$align-hours-arrow: ($width-hours-arrow / 2) - $width-hours-arrow;
+$align-seconds-arrow: ($width-seconds-arrow / 2) - $width-seconds-arrow;
+
+$radius: 50%;
+$watch-border-width: $size-container-watch * 0.02;
+
+$border-color: #ff2e63;
+$circle-bg-color: #393e46;
+$seconds-bg-color: #71c9ce;
+$hours-bg-color: #ff9a00;
```
StasSokolov, line 16: create `scss` variables when you use value more than one time

StasSokolov (changes requested):
Good job, let's improve your code a little more!

StasSokolov (approved):
Great job🔥

PR #2740, opened 2023-09-05
add task

AlinaChukh:


src/styles/container.scss
```diff
@@ -0,0 +1,49 @@
...
+  border-radius: $border-radius;
+
+  &__elem {
+    display: flex;
+    margin: auto;
+
+    &--circle {
+      position: absolute;
+      z-index: 1;
+      background-color: orange;
```
OleziO, line 18: Use HEX instead of color names.

OleziO (changes requested):
Fix demo link
image

src/styles/container.scss
```diff
@@ -0,0 +1,51 @@
...
+    &--little {
+      background-color: #00f;
+      height: 4%;
+      position: absolute;
+      left: 50%;
+      top: 48%;
+      width: 25%;
+      transition: transform 60s;
+      transform-origin: left;
+      transform: rotate(360deg);
```
TarasHoliuk, line 48: you should use `animation`, not `transition`
Currently it's just a static image

The transition property doesn't cause any "moves" just by itself.

TarasHoliuk (changes requested):

src/styles/container.scss
```diff
@@ -0,0 +1,64 @@
+$content-width: 600px;
+$border-radius: 50%;
```
DarkMistyRoom, line 2: would be good to move variables to a separate file

src/styles/container.scss
```diff
@@ -0,0 +1,64 @@
...
+
+@keyframes run {
+  0% {
+    transform: rotate(0);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
DarkMistyRoom, line 11: same for keyframes

src/styles/container.scss
```diff
@@ -0,0 +1,64 @@
...
+      position: absolute;
+      left: 50%;
+      top: 48%;
+      width: 25%;
+      animation-name: run;
+      transform-origin: left;
+      animation-duration: 60s;
+      animation-iteration-count: infinite;
+      animation-play-state: running;
+    }
```
DarkMistyRoom, line 62: avoid styles duplication. Use common class or mixin/extend

DarkMistyRoom (changes requested):
As I remember the concept of clock, the second hand should make the circle in 60s and minute one in 60m with linear speed. Your clock breaks this concept) It's preety originally, but let's make a classic variant of clock

etojeDenys (approved):
lgtm

PR #2741, opened 2023-09-05
add task solution

pzwolenik:
- [DEMO LINK](https://pzwolenik.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,49 @@
+@import "../utils/variables";
+
+.clock {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  box-sizing: border-box;
+  width: $clock-dim;
+  height: $clock-dim;
+  border: 5px dotted black;
+  border-radius: 50%;
+
+  &__center {
+    position: absolute;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,49 @@
+@import "../utils/variables";
+
+.clock {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  box-sizing: border-box;
+  width: $clock-dim;
+  height: $clock-dim;
+  border: 5px dotted black;
```
OleziO, line 10: Use HEX instead of color names. Color names can vary in rendering between browsers and devices. Hexadecimal values are more consistent and predictable across different platforms, ensuring that your chosen color appears as intended.

OleziO: Check other similar cases.

OleziO (changes requested):

src/styles/blocks/clock.scss
pzwolenik: Thanks! :)

plodnikate (approved):

PR #2742, opened 2023-09-05
Develop

HokageHEL:
- [DEMO LINK](https://HokageHEL.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/index.scss
+++ b/src/styles/blocks/index.scss
@@ -1,3 +1,7 @@
+* {
+  box-sizing: border-box;
+}
+
 body {
   margin: 0;
   height: 80vmin;
--- a/src/styles/blocks/index.scss
+++ b/src/styles/blocks/index.scss
@@ -1,0 +1,6 @@
+body {
+  margin: 0;
+  height: 80vmin;
+
+  @include content-center;
+}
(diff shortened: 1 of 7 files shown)


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,47 @@
...
+
+  @function duration($time) {
+    @return $time/60;
+  }
+
+  position: relative;
+  width: $stopwatch-size;
+  height: $stopwatch-size;
+  border-radius: 50%;
+  border: 2px dotted black;
```
OleziO, line 12: Use HEX instead of color names. Color names can vary in rendering between browsers and devices. Hexadecimal values are more consistent and predictable across different platforms, ensuring that your chosen color appears as intended.

OleziO (changes requested):
Fix anumation (For the demo purpose rotate hands during 5s and 60s)
image

src/styles/blocks/index.scss
```diff
@@ -0,0 +1,10 @@
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  height: 80vmin;
```
maxim2310, line 7: its mean size of the clock, not body)
image

HokageHEL: ah ok sry
ty!

PR #2751, opened 2023-09-08
add task solution

PaninDmytro:
https://PaninDmytro.github.io/layout_stop-watch/

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -8,7 +8,11 @@
     <title>Stop watch</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Stop watch</h1>
+  <body class="body">
+    <div class="circle body__circle">
+      <div class="circle__firstHand"></div>
+      <div class="circle__secondHand"></div>
+      <div class="circle__centre"></div>
+    </div>
   </body>
 </html>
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,59 @@
-// put code here
+$circleSize: 80vmin;
```
tilugdmytro, line 1: You could make some more variables and move it to the utils folder, as well as create block folder with circle.scss.
main.scss in that case would contain only imports. Just for your practice 😃

src/styles/main.scss
```diff
@@ -1 +1,59 @@
...
+
+body {
+  margin: 0;
```
tilugdmytro, line 4: It is not very convenient to look at the watch because its at the very top of the page, add some space please.

src/index.html
```diff
@@ -8,7 +8,11 @@
     <title>Stop watch</title>
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
-  <body>
-    <h1>Stop watch</h1>
+  <body class="body">
+    <div class="circle body__circle">
```
tilugdmytro, line 12: [Checklist 5] Wrong BEM structure. You don't need body__circle class. To center your watch just add "margin: 0 auto" to the body tag

tilugdmytro (changes requested):
Great job overall! Please, add some changes.

vasylmiskiv (approved):

PR #2759, opened 2023-09-10
Completed task

AnonimToster:


src/styles/main.scss
```diff
@@ -1 +1,66 @@
 // put code here
+$watch-size: 80vmin;
```
polosanya, line 2: It would be great to create a separate file for variables

src/styles/main.scss
```diff
@@ -1 +1,66 @@
...
+
+.watch {
```
polosanya, line 4: Same with BEM blocks

src/styles/main.scss
```diff
@@ -1 +1,66 @@
...
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+    animation-duration: 5s;
+    animation-name: rotating;
+  }
+
+  &__minutes {
+    position: absolute;
+    top: #{$watch-size / 2 - $watch-size * 0.25};
+    left: #{$watch-size / 2 - $watch-size / 48};
```
polosanya, line 32: image
Instead of hardcoding the exact sizes and distances, you can just center them with `transform: translate`
https://www.w3schools.com/css/tryit.asp?filename=trycss_align_transform

polosanya (changes requested):
Almost perfect!

src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,65 @@
...
+    height: #{$watch-size / $center-radius-ratio};
+    transform: translate(-50%, -50%);
+    background-color: yellow;
+    border-radius: 50%;
+  }
+}
+
+@keyframes rotating {
+  0% {
+    transform: translate(-50%, -100%) rotate(0deg);
```
polosanya, line 55: You can change transform just in styles, no need to duplicate it in animation

polosanya (approved):
👍

PR #2764, opened 2023-09-12
add stopwatch

IwonaIvona:
[DEMO LINK](https://IwonaIvona.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/stopwatch.scss
+++ b/src/styles/blocks/stopwatch.scss
@@ -1,0 +1,54 @@
+.stopwatch {
+  box-sizing: border-box;
+  position: relative;
+  width: $stopwatchSize;
+  height: $stopwatchSize;
+  justify-content: center;
+  align-items: center;
+  
+  &--face {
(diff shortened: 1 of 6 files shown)


src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,54 @@
+.stopwatch {
+  box-sizing: border-box;
+  position: relative;
+  width: $stopwatchSize;
+  height: $stopwatchSize;
+  justify-content: center;
+  align-items: center;
+  
+  &--face {
```
Centmsn, line 9: The modifier is not really needed here since we are not using the unmodified version of the `stopwatch` class. We could include those rules inside `stopwatch` class itself.

Centmsn (approved):
Great 🚀 

I left You a single comment to consider 😃

src/styles/blocks/stopwatch.scss
IwonaIvona: Yes, that's right :) I didn't notice at the end that it was unnecessary. Thank you :)

mykhalenych (approved):

PR #2772, opened 2023-09-15
add task solution

viktalfrontend:


maxim2310 (changes requested):
need to fix according to requirements of task 
image

sTorba24 (approved):

PR #2775, opened 2023-09-18
add task solution

rdudchar:


src/styles/style.scss
```diff
@@ -0,0 +1,71 @@
...
+}
+
+@keyframes rotateSecondHand {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
```
VitaliyBodnar, line 71: these two keyframes are same, so don't need to duplicated

VitaliyBodnar (changes requested):

etojeDenys (approved):
lgtm

PR #2778, opened 2023-09-19
add task solution

Kalkorejz:
- [DEMO LINK](https://Kalkorejz.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,53 @@
+.watch {
+  position: relative;
+  margin: auto;
+  top: 10vmin ;
+
+  height: 80vmin;
+  width: 80vmin;
+
+  border: 1.7vmin dotted black;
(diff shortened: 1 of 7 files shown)


src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,53 @@
...
+
+  &__hours {
+    position: absolute;
+    left: (100% - $hoursWidth)/2;
+    top: (50% - $hoursHeight);
+
+    width: $hoursWidth;
+    height: $hoursHeight;
+
+    background: blue;
```
OleziO, line 34: Try to use HEX instead of color names (because the colors given by the names will vary depending on the browser)

OleziO (changes requested):

OleziO (approved):

Kalkorejz:
tywm!

PR #2782, opened 2023-09-20
add task solution

wassabi7:
- [DEMO LINK](https://wassabi7.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/hand.scss
+++ b/src/styles/blocks/hand.scss
@@ -1,0 +1,55 @@
+.hand {
+  box-sizing: border-box;
+  transform-origin: bottom;
+  position: absolute;
+  left: 50%;
+  transform: translateX(-50%);
+
+  &__minute {
+    width: $minuteWidth;
+    height: $minuteHeight;
+
+    border: 3px solid #89CFEF;
+
+    top: 50% - $minuteHeight;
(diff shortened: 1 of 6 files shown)


src/styles/blocks/hand.scss
```diff
@@ -0,0 +1,55 @@
...
+
+    width: $pointWidth;
+    height: $pointWidth;
+
+    position: absolute;
+
+    top: 50%;
+    left: 50%;
+
+    transform: translate(-50%, -50%);
```
VitaliyBodnar, line 44: Let's try to organize your CSS
I recommend you take a look at these styles code and use it in the feature
[9elements](https://9elements.com/css-rule-order) or [htmlAcademy](https://codeguide.academy/html-css.html#css-order)

VitaliyBodnar (changes requested):
- these are few mistakes with BEM, blocks elements should be inside to block
Screenshot 2023-09-20 at 15 33 25

src/styles/blocks/stopwatch.scss
```diff
@@ -0,0 +1,16 @@
+.stopwatch {
+  position: relative;
```
Andriy-Shymkiv, line 2: you don't use it, so let's remove
```suggestion
```

PR #2790, opened 2023-09-23
Beta watch

Stan1slav123:


src/styles/main.scss
```diff
@@ -1 +1,177 @@
-// put code here
+@keyframes minuteAnimated {
+  0% {
```
VitaliyBodnar, line 2: instead of using percents you can use just `from` and `to`

src/styles/main.scss
```diff
@@ -1 +1,177 @@
...
+
+  position: absolute;
+  height: 40px;
+  width: 40px;
+  border-radius: 50%;
+  background-color: #ff0;
+  top: 50%;
+  left: 50%;
+  transform: translate(-25%, -50%);
+}
```
VitaliyBodnar, line 153: you need to use scss structure, for instance:
```suggestion
.watch {
  width: 80vmin;
  height: 80vmin;
  border-radius: 50%;
  border: 10px dotted #000;
  
  &__center {
    position: absolute;
    height: 40px;
    width: 40px;
    border-radius: 50%;
    background-color: #ff0;
    top: 50%;
    left: 50%;
    transform: translate(-25%, -50%);
  }
}
```
check other places

VitaliyBodnar (changes requested):

TarasHoliuk (approved):

PR #2799, opened 2023-09-29
task done, check please

olehpushkar:
[DEMO LINK](https://olehpushkar.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/stopwatch.scss
+++ b/src/styles/stopwatch.scss
@@ -1,0 +1,69 @@
+.stopwatch {
+  position: relative;
+  box-sizing: border-box;
+
+  width: 80vmin;
+  height: 80vmin;
+
+  border: 5px dotted;
+  border-radius: $watchBorderRadius;
+
+  @include with-centered-content;
+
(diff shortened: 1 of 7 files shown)


maxim2310 (changes requested):
Clock must go anytime not just on hover

olehpushkar:
thanks, fixed it

src/styles/stopwatch.scss
```diff
@@ -0,0 +1,63 @@
...
+
+    animation: $watchAnimation;
+    animation-duration: 5s;
+  }
+}
+
+@keyframes roll {
+  0% {
+    transform: rotate(0);
+    transform-origin: $watchTransformOrigin;
```
etojeDenys, line 56: why do we need this property if we never change it?

etojeDenys (approved):
good job

src/styles/stopwatch.scss
olehpushkar: Thank you for checking!

PR #2810, opened 2023-10-03
add task solution

DudeMood:
https://DudeMood.github.io/layout_stop-watch/


lerastarynets (approved):
Such an unusual implementation of the task. However, I think if u were able to achieve what you've made, then implementing the original variant of the clock should be easy for you. So really really well done, I adore creative approach 🔥

PR #2825, opened 2023-10-10
stop watch

JuliaKnysh:
[DEMO LINK](https://JuliaKnysh.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,20 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Stop watch</h1>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -9,6 +9,20 @@
...
+
+
+
+
+
+
+
+
+
+
```
volodymyr-soltys97, line 28:
```suggestion

```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your solution
1. It is necessary to correct the time for which the arrows make a full revolution according to the description of the task
image

PR #2827, opened 2023-10-12
add task solution

AlinaMyronenko:
https://AlinaMyronenko.github.io/layout_stop-watch

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,59 @@
-// put code here
+body {
+  display: flex;
+  justify-content: center;
+  margin: 0;
+}
+
+.stopwatch {
+  position: relative;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,59 @@
...
+  border-radius: 50%;
+  border: 3px dotted red;
+
+  &:before {
+    content: "";
+    position: absolute;
+    width: 30px;
+    height: 30px;
+    border-radius: 50%;
+    background-color: red;
```
VitaliyBodnar, line 25: red color is duplicated, so make sense to move it to the variable

VitaliyBodnar (changes requested):
Almost done!

PR #2834, opened 2023-10-14
add task solution

TetianaBukhenko:
- [DEMO LINK](https://TetianaBukhenko.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,60 @@
-// put code here
+// // put code here
+body {
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stop-watch {
+  &_container {
+    display: flex;
+    justify-content: center;
+    align-items: flex-end;
+    border: 5px solid black;
+    border-radius: 50%;
+    height: 500px;
+    width: 500px;
+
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,60 @@
...
+      border: 5px solid #212020;
+      border-radius: 25%;
+      background-color: #5b5a5a4a;
+      height: 150px;
+      width: 20px;
+      margin-bottom: 50%;
+      z-index: 0;
+      animation: rotate 60s linear 7s infinite;
+      transform-origin: bottom;
+    }
```
VitaliyBondarenko1982, line 52: hand-big and hand-small has same properties - suggest to make mixin for them with arguments - values of properties which different.

VitaliyBondarenko1982 (changes requested):
make clock more realistic
on start arrows should be in same position/
expected:
Screenshot from 2023-10-14 18-49-05

actual:
Screenshot from 2023-10-14 18-49-33

TetianaBukhenko:
Hello, please check an updated task solution.

VitaliyBondarenko1982 (approved):
Well done. Much better)

PR #2844, opened 2023-10-17
created stopwatch according to the tasks assigned

Oleksandr-Nahirnyi:
- [DEMO LINK](https://Oleksandr-Nahirnyi.github.io/layout_stop-watch/)

-x Keyframes implemented using from/to + transform with rotate property
 -x Stopwatch is centered and has correct arrows size
 -x All Typical Mistakes from BEM lesson theory are checked.
 -x Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,57 @@
-// put code here
+.body {
+  width: 100%;
+  height: 100vh;
+  display: grid;
+  place-items: center;
+  background-color: black;
+}
+
+.stopwatch {
+  position: relative;
+  width: 500px;
+  height: 500px;
+  border: 15px dotted darkred;
+  border-radius: 50%;
+
+  &__center-button {
+    position: absolute;
+    top: 235px;
+    left: 235px;
+    width: 30px;
+    height: 30px;
+    border-radius: 50%;
+    background-color: yellow;
+  }
+
+  &__minute-hand {
+    position: absolute;
+    top: 0;
+    left: 245px;
+    height: 250px;
+    width: 10px;
+    background-color: blue;
+    transform-origin: bottom center;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,57 @@
...
+  position: relative;
+  width: 500px;
+  height: 500px;
+  border: 15px dotted darkred;
+  border-radius: 50%;
+
+  &__center-button {
+    position: absolute;
+    top: 235px;
+    left: 235px;
```
anastasiia-tilikina, line 19: it is better to use relative points of measure to center this element. You can make it like this:

```scss
top: 50%;
left: 50%;
transform: translate(-50%, -50%);
```

src/styles/main.scss
```diff
@@ -1 +1,57 @@
...
+  &__second-hand {
+    position: absolute;
+    top: 100px;
+    left: 240px;
+    height: 150px;
+    width: 20px;
+    background-color: blue;
+    transform-origin: bottom center;
+    animation: rotate 60s infinite steps(60);
+  }
```
anastasiia-tilikina, line 46: you have some duplicates here,
create a common class for instance `&__hand`, and set styles by default. and for second-hand override that you need

anastasiia-tilikina (changes requested):
Good job!

check my comments.

Pay attention to this point 
image

anastasiia-tilikina (changes requested):
seems like you haven't redeploy your project
I see some changes in the code but they haven't applied in your demo

Esceype (approved):
GJ!

PR #2853, opened 2023-10-19
Develop

yevtok:
- [DEMO LINK](https://yevtok.github.io/layout_stop-watch/)


src/styles/anim.scss
ArtemTeslenko, line 1: is it really so difficult to write 'animation'?😁 file's names should be meaningful

src/styles/clock.scss
```diff
@@ -0,0 +1,47 @@
...
+  border-radius: 50%;
+
+  &__min-hand {
+    position: absolute;
+    z-index: 1;
+    background-color: $color-min;
+    border: 2px solid #000;
+    width: $min-hand-width;
+    height: $clock-size / 4;
+    left: calc(50% - #{$min-hand-width} / 2);
```
ArtemTeslenko, line 15: to avoid this calculations and make element's behavior more predictable you can use pattern which center your element:

> top: 50%;
>   left: 50%;
>   transform: translate(-50%, -50%);

src/styles/vars.scss
ArtemTeslenko, line 1: variables

ArtemTeslenko (approved):
Good) i am approving it, but see comments

PR #2861, opened 2023-10-24
code clock

vitya2305:
- [DEMO LINK](https://vitya2305.github.io/layout_stop-watch/)

- [+] Keyframes implemented using from/to + transform with rotate property
- [+] Stopwatch is centered and has correct arrows size
- [+] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [+] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,113 @@
-// put code here
+
+html {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100%;
+}
+
+.clock {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  position: relative;
+  width: 500px;
+  height: 500px;
+
+  border-radius: 50%;
+
+  &__second-hand {
+    border-radius: 5px;
+    position: absolute;
+
+    transform: translateY(-50%);
+
+    height: 50%;
+    width: 8px;
+    background-color: rgb(253, 200, 20);
+    z-index: 6;
+
+    animation: rotateHand 60s linear infinite;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,113 @@
...
+  }
+
+  &__minute-hand {
+    height: 25%;
+    width: 15px;
+    position: absolute;
+    background-color: rgb(191, 209, 130);
+    z-index: 5;
+    
+    animation: rotateHand 3600s linear infinite;
```
volodymyr-soltys97, line 41: Need to fix the time for rotate hands as on the task description
image

src/styles/main.scss
```diff
@@ -1 +1,113 @@
...
+}
+
+.clock {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
+  position: relative;
+  width: 500px;
+  height: 500px;
```
volodymyr-soltys97, line 16: Need to fix
image

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your solution

src/styles/main.scss
vitya2305: If I understood correctly instruction, instead of 60 and 3600s for seconds-hand and minute-hand, should I use 5s and 60s?

volodymyr-soltys97 (approved):
Well done 🔥

PR #2869, opened 2023-10-27
ready

DanyloSemenov:
https://DanyloSemenov.github.io/layout_stop-watch/
 [x ] Keyframes implemented using from/to + transform with rotate property


mvjl000 (changes requested):
Animations should be infinite. The bigger hand should make a full circle in 5s, the smaller one in 60s

mvjl000 (approved):
Good job! 🚀

PR #2876, opened 2023-10-31
add task solution

vneholiuk:
- [DEMO LINK](https://vneholiuk.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/hand.scss
+++ b/src/styles/hand.scss
@@ -1,0 +1,43 @@
+.hand {
+  position: absolute;
+  top: 0;
+  left: 49%;
+
+  transform-origin: center bottom;
+  transform: translate(-50%, -50%);
+
+  &--minute {
+    top: 20%;
+    width: 2 * $hand-size;
+    height: $hand-height;
+    background-color: $hand-minute-color;
+    animation: rotateMinuteHand 60s infinite linear;
+    border-radius: 10px;
+  }
+
+  &--second {
+    width: $hand-size;
+    height: $hand-height + 20%;
+    background-color: $hand-second-color;
+    animation: rotateSecondHand 5s infinite linear;
+    border-radius: 5px;
+  }
+}
+
+@keyframes rotateMinuteHand {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
+
+@keyframes rotateSecondHand {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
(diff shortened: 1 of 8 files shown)


nazarmatsevych (changes requested):
Good work, but update your styles due to my screen to remove scroll
image

src/styles/hand.scss
```diff
@@ -0,0 +1,43 @@
...
+}
+
+@keyframes rotateSecondHand {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
```
volodymyr-soltys97, line 43: These 2 animations do the same thing, so you need to remove one

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review
image

src/styles/hand.scss
anastasiia-tilikina: not fixed

anastasiia-tilikina (changes requested):
please fix issue from previous review

src/styles/hand.scss
vneholiuk: this animations do the same thing but if I will remove one, one of them will not perform animation

vneholiuk: the first makes a full revolution in 60 seconds, and the second makes a revolution in 5 seconds

lerastarynets: @vneholiuk this isn't true, you don't set timing in this animations, u can just leave one of them and call it `rotateHand` and then in `&--second` you make `animation: rotateHand 5s infinite linear;` and in `&--minute` you do `animation: rotateHand 60s infinite linear;`

lerastarynets (changes requested):
Pls fix prev comments

src/styles/main.scss
```diff
@@ -1 +1,5 @@
-// put code here
+@import "../../src/styles/variables.scss";
+@import "../../src/styles/reset.scss";
+@import "../../src/styles/stopwatch.scss";
+@import "../../src/styles/hand.scss";
+@import "../../src/styles/center.scss";
```
volodymyr-soltys97, line 5:
```suggestion
@import "./variables.scss";
@import "./reset.scss";
@import "./stopwatch.scss";
@import "./hand.scss";
@import "./center.scss";
```

volodymyr-soltys97 (changes requested):
Almost done!

volodymyr-soltys97 (approved):
Well done 🔥

PR #2885, opened 2023-11-05
add task solution

Val-Bryks:
- [DEMO LINK](https://Val-Bryks.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/arrows.scss
+++ b/src/styles/blocks/arrows.scss
@@ -1,0 +1,42 @@
+.arrow {
+  position: absolute;
+  top: 0;
+
+  transform-origin: center bottom;
+  transform: translate(-50%, -50%);
+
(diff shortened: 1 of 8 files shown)


src/styles/blocks/arrows.scss
```diff
@@ -0,0 +1,42 @@
...
+}
+
+@keyframes short-arrow-move {
+  0% {
+    transform: rotate(0deg);
+  }
+  100% {
+    transform: rotate(360deg);
+  }
+}
```
nazarmatsevych, line 42: You have two identical animations, so you can only use one in both cases

nazarmatsevych (approved):
Good work, left minor comment

maxim2310 (approved):
Good work!!!

PR #2888, opened 2023-11-07
add task solution

SerjDSA:
- [DEMO LINK](https://serjdsa.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -9,6 +9,15 @@
     <link rel="stylesheet" href="./styles/main.scss">
   </head>
   <body>
-    <h1>Stop watch</h1>
+    <div class="container">
+      <div class="stopwatch">
+        <div class="stopwatch__hand stopwatch__hand--min">
+          <div class="stopwatch__hand-arrow stopwatch__hand-arrow--min"></div>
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -9,6 +9,15 @@
...
+    <div class="container">
+      <div class="stopwatch">
+        <div class="stopwatch__hand stopwatch__hand--min">
+          <div class="stopwatch__hand-arrow stopwatch__hand-arrow--min"></div>
+        </div>
+        <div class="stopwatch__hand stopwatch__hand--sec">
+          <div class="stopwatch__hand-arrow stopwatch__hand-arrow--sec"></div>
+        </div>
+      </div>
+    </div>
```
denys-danyliuk, line 21: do you need so many elements?
you could use 3/4 and it would work the same as now

denys-danyliuk (approved):
GJ!

src/index.html
SerjDSA: Good afternoon. Without a container it will not center. In body not woriking as I want. I started with three div..

PR #2892, opened 2023-11-09
add task solution

alshust:
[DEMO LINK](https://alshust.github.io/layout_stop-watch/)


nazarmatsevych (changes requested):
Great work but let's fix your animation to be linear

TarasHoliuk (approved):
GJ :+1: Approved with one comment - each BEM block should be in a separate file (.stopwatch in this case)

No need to re-request review - already approved, but take into account in the future

PR #2908, opened 2023-11-16
add task solution

OksanaNastenko:

### DIFF
--- a/src/styles/blocks/clock.scss
+++ b/src/styles/blocks/clock.scss
@@ -1,0 +1,62 @@
+* {
+  margin: 0;
+  padding: 0;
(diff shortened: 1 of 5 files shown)


src/styles/blocks/clock.scss
```diff
@@ -0,0 +1,62 @@
+* {
+  margin: 0;
+  padding: 0;
```
TarasHoliuk, line 3: it's bad practice to reset such values in the `*` selector. The only thing allowed to set here is `box-sizing`

TarasHoliuk (approved):
LGTM :+1: 
Added 1 comment FYI

PR #2912, opened 2023-11-17
add task solution

Svimpik:
- [DEMO LINK](https://Svimpik.github.io/layout_stop-watch/)

 [x] Keyframes implemented using from/to + transform with rotate property

### DIFF
--- a/src/styles/block-style/page-style.scss
+++ b/src/styles/block-style/page-style.scss
@@ -1,0 +1,11 @@
+* {
+  margin: 0;
+  box-sizing: border-box;
+}
+
+body {
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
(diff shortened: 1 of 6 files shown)


src/styles/block-style/stopwatch-style.scss
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  @keyframes rotate-minute {
+    0% {
+      transform: translateX(-50%) rotate(0deg);
+    }
+    100% {
+      transform: translateX(-50%) rotate(360deg);
+    }
+  }
```
nazarmatsevych, line 54: what for you need two identical animations? Remove one of them

src/styles/block-style/page-style.scss
```diff
@@ -0,0 +1,11 @@
+* {
+  margin: 0;
```
nazarmatsevych, line 2: Don't reset margin or padding in * selector, reset only for elements you are using

nazarmatsevych (changes requested):
Update this part pls
image

nazarmatsevych (approved):
Well done 👍

PR #2917, opened 2023-11-22
Job's done

AndrewEdgers:
- [DEMO LINK](https://AndrewEdgers.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,70 @@
-// put code here
+$clock-size: 80vmin;
+
+@mixin center {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+}
+
+@mixin hands-position {
+  transform-origin: 50% 100%;
+  transform: rotate(0deg);
+}
+
+body {
+  margin: 0;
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,70 @@
...
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+}
+
+@mixin hands-position {
+  transform-origin: 50% 100%;
+  transform: rotate(0deg);
+}
```
maxim2310, line 13: move it to separate files

maxim2310 (changes requested):
Great job overall! Just couple tiny things needs to be fixed before approval)
something wrong in youк clock face
image

src/styles/main.scss
AndrewEdgers: Clock face was tested on 1920x1080 screen only 😂, it's only border and that was half intentional. Will do


> move it to separate files

Should I do it even if such small projects? Or is it as a habit/sake of task?

anastasiiavorobiova (approved):
Great work!

PR #2921, opened 2023-11-26
task solution

hsvirina:
https://hsvirina.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -11,6 +11,7 @@ body {
   border: dotted black;
   border-radius: 50%;
   position: relative;
+  box-sizing: border-box;
 
   &::before {
     content: "";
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,61 @@
-// put code here
+body {
+  display: flex;
+  height: 100vh;
+
```
VitaliyBodnar, line 4:
```suggestion
```

src/styles/main.scss
```diff
@@ -1 +1,61 @@
...
+}
+
+.watch {
+  margin: auto;
+  height: 80vmin;
+  width: 80vmin;
+  border: dotted black;
+  border-radius: 50%;
+  position: relative;
+  box-sizing: border-box;
```
VitaliyBodnar, line 14: It's better to move box sizing to this block `*{}` as it's better practice

VitaliyBodnar (changes requested):

volodymyr-soltys97 (approved):
Great work 🔥

PR #2941, opened 2023-12-03
'stop-watch task solution'

DenLysiak:


src/styles/keyframes.scss
```diff
@@ -0,0 +1,19 @@
...
+
+@keyframes move-minute {
+  from {
+    transform: rotate(90deg);
+  }
+
+  to {
+    transform: rotate(450deg);
+  }
+}
```
volodymyr-soltys97, line 19: You can use only one keyframes for rotate

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
Need to fix time rotate
image

VitaliyBodnar (approved):
Good job!

PR #2956, opened 2023-12-09
Develop

manch0ffline:
- [DEMO LINK](https://manch0ffline.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,5 +1,4 @@
-@import 'variables.scss';
-
+@import 'variables';
 
 * {
   box-sizing: border-box;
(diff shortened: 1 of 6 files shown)


src/styles/main.scss
```diff
@@ -1 +1,77 @@
...
+* {
+  box-sizing: border-box;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+
+  display: flex;
+  flex-direction: column;
```
l4st1m0za, line 12: The flex-direction property is set to column, which is not necessary in this case as there's only one child element.

l4st1m0za (approved):
Awesome job!

PR #3012, opened 2024-01-08
add task solution

yuriiyepikhov:
- [DEMO LINK](https://yuriiyepikhov.github.io/layout_stop-watch/)
- [x] Code follows all the Code Style Rules

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,83 @@
 // put code here
+
+$time: 45;
+
+.page {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  width: 100vw;
+  height: 100vh;
+  margin: 0;
+}
+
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,83 @@
...
+
+@keyframes small-arrow-move {
+  from {
+    transform: rotate(0);
+  }
+
+  to {
+    transform: rotate(#{$time / 3600 * 360}deg);
+  }
+}
```
maxim2310, line 83: you dont need 2 keyframes for same circle rotation, and now its finite but must be not

maxim2310 (changes requested):
Good work, but lets make some improvement

anastasiia-tilikina (approved):
Well done!

PR #3016, opened 2024-01-09
add task solution

kredynetsA:
- [DEMO LINK](https://kredynetsa.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,47 @@
-// put code here
+@keyframes rotate-minute-hand {
+  from { transform: rotate(0deg); }
+  to { transform: rotate(360deg); }
+}
+
+@keyframes rotate-hour-hand {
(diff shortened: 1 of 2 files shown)


src/styles/main.scss
```diff
@@ -1 +1,47 @@
-// put code here
+@keyframes rotate-minute-hand {
+  from { transform: rotate(0deg); }
+  to { transform: rotate(360deg); }
+}
+
+@keyframes rotate-hour-hand {
+  from { transform: rotate(0deg); }
+  to { transform: rotate(360deg); }
+}
```
lerastarynets, line 9: this animations are the same, you can just have 1 animation and use it both places

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

anastasiiavorobiova (approved):
Awesome work!

PR #3019, opened 2024-01-10
stopwatch 

Maryna1313:
- [DEMO LINK](https://Maryna1313.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -17,7 +17,6 @@
     width: 20px;
     border-radius: 50%;
     z-index: 2;
-
   }
 
   &__hand-sec,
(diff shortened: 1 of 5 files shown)


src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,54 @@
...
+  &__hand-sec,
+  &__hand-min {
+    position: absolute;
+    z-index: 1;
+    transform-origin: 50% 100%;
+
+    animation-name: move;
+    animation-timing-function: linear;
+    animation-iteration-count: infinite;
+    animation-fill-mode: backwards;
```
maxim2310, line 31: would be better to combine it in one animation rule 
image

maxim2310 (approved):
Good work!!!

PR #3026, opened 2024-01-14
add layout watch

self51:
- [DEMO LINK](https://self51.github.io/layout_stop-watch/)


nazarmatsevych (changes requested):
Animation should be linear and infinite
image
and long arrow should be faster than short one

VitaliyBondarenko1982 (approved):
Looks good

PR #3027, opened 2024-01-14
add new design Apple Watch 10

vhrubyi:
- [DEMO LINK](https://vhrubyi.github.io/layout_stop-watch/)

- [x] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/src/styles/blocks/watch.scss
+++ b/src/styles/blocks/watch.scss
@@ -1,0 +1,64 @@
+.watch {
+  display: flex;
+  height: 80vmin;
+  width: 80vmin;
+  border: 2px;
+  border-radius: 50%;
+  border-style: dotted;
+  position: relative;
+  align-items: center;
+  justify-content: center;
+
+  &__second {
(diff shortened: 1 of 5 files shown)


src/styles/blocks/watch.scss
```diff
@@ -0,0 +1,62 @@
...
+}
+
+@keyframes rotate-minute {
+  0% {
+    transform: translate(-50%, -100%) rotate(0deg);
+  }
+  100% {
+    transform: translate(-50%, -100%) rotate(360deg);
+  }
+}
```
SanyaBratashchuk, line 62: why do you need two identical keyframes it can be just one

SanyaBratashchuk: also it is better to create a separate file for keyframes

SanyaBratashchuk (changes requested):

TarasHoliuk (approved):

PR #3046, opened 2024-01-22
add solution

MatheusStefan:
- [DEMO LINK] (https://MatheusStefan.github.io/layout_stop-watch/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1368,9 +1368,9 @@
       "dev": true
     },
     "@mate-academy/scripts": {
-      "version": "1.7.2",
-      "resolved": "https://registry.npmjs.org/@mate-academy/scripts/-/scripts-1.7.2.tgz",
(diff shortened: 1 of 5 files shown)


src/styles/main.scss
```diff
@@ -1 +1,86 @@
...
+
+    &__hour {
+        position: absolute;
+        left: 0;
+        right: 0;
+        margin: 0 auto;
+        width: 30px;
+        height: 100%;
+        z-index: 1;
+        animation: time 5s linear;
```
jv-aquino, line 35: The stopwatch should be animated infinitely and you can implement it very easily with css!

jv-aquino (changes requested):
Your code is great and you are using BEM in a very good way! I would just ask to correct this one thing from the checklist and also the markdown is not formated well. Keep up your good work!

pedro-ruas (approved):
Great Job!

PR #3047, opened 2024-01-22
add task solution

pakhomovalex:
[DEMO LINK](https://pakhomovalex.github.io/layout_stop-watch/)
 
### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,60 @@
 // put code here
+@import "./variables";
+
+.display {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  box-sizing: border-box;
+  width: $box-size;
+  height: $box-size;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border: $border-size dotted;
+  border-radius: 50%;
+}
+
+.display_center {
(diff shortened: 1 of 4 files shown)


src/styles/main.scss
```diff
@@ -1 +1,60 @@
...
+
+.display_center {
+  width: $center-point-size;
+  height: $center-point-size;
+  border-radius: 50%;
+  background-color: red;
+  z-index: 1;
+}
+
+.display_seconds-hand {
```
l4st1m0za, line 27: The class naming does not follow BEM (Block Element Modifier) methodology. The class should be named as 'display__seconds-hand' instead of 'display_seconds-hand'.

src/styles/main.scss
```diff
@@ -1 +1,60 @@
...
+  width: $box-size;
+  height: $box-size;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  border: $border-size dotted;
+  border-radius: 50%;
+}
+
+.display_center {
```
l4st1m0za, line 19: The class naming does not follow BEM (Block Element Modifier) methodology. The class should be named as 'display__center' instead of 'display_center'.

l4st1m0za (approved):
Good job, just change naming for the BEM

PR #3049, opened 2024-01-23
implement solution

nester7an:
- [DEMO LINK](https://nester7an.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/utilities/variables.scss
+++ b/src/styles/utilities/variables.scss
@@ -1,0 +1,8 @@
+$clock-size: 400px;
+$border-size: 10px;
+$hand-min-width: 5px;
+$hand-hour-width: 8px;
+$circle-size: 25px;
+
+
+
--- a/readme.md
+++ b/readme.md
@@ -15,7 +15,7 @@ Create a working stopwatch with minute and second hands using only CSS animation
 ## Checklist
 
 ❗️ Replace `<your_account>` with your Github username and copy the links to `Pull Request` description:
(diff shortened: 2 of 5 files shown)


src/styles/utilities/variables.scss
```diff
@@ -0,0 +1,8 @@
+$clock-size: 400px;
+$border-size: 10px;
+$hand-min-width: 5px;
+$hand-hour-width: 8px;
+$circle-size: 25px;
+
+
+
```
anastasiiavorobiova, line 8:
```suggestion

```

anastasiiavorobiova (changes requested):
Great work! Consider implementing these task requirements
Screenshot 2024-01-23 at 14 48 20

src/styles/utilities/variables.scss
```diff
@@ -0,0 +1,8 @@
+$clock-size: 400px;
```
volodymyr-soltys97, line 1: Do you use this variable?
```suggestion
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
The clock needs to be centered on the page vertically and horizontally, also now the hands move only when we move the cursor, this is wrong and the hands should move constantly

Esceype (changes requested):
GJ!
But the stopwatch must have animation by default state not only when we move the mouse on the stopwatch block(hover state).

vadiimvooo (approved):
Well done!)

PR #3050, opened 2024-01-24
add task solution

arthur10031198:
- [DEMO LINK](https://arthur10031198.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,66 @@
-// put code here
+  body {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 100vh;
+    margin: 0;
+  }
+
+    .stopwatch {
+      background-color: #333;
+      position: relative;
+      width: 80vmin;
+      height: 80vmin;
+      border: 4px dashed gold;
+      border-radius: 50%;
+      overflow: hidden;
+    }
+
+    .stopwatch__hand {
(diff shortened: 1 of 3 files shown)


src/styles/main.scss
```diff
@@ -1 +1,66 @@
...
+
+    @keyframes rotate-second-hand {
+      0% {
+        transform: translateX(-50%) translateY(-100%) rotate(-90deg);
+      }
+
+      100% {
+        transform: translateX(-50%) translateY(-100%) rotate(270deg);
+      }
+    }
```
lerastarynets, line 66: these animations are the same, just leave one and call it `rotate-hand`, and use where needed

src/styles/main.scss
```diff
@@ -1 +1,66 @@
...
+      background-color: #333;
+      position: relative;
+      width: 80vmin;
+      height: 80vmin;
+      border: 4px dashed gold;
+      border-radius: 50%;
+      overflow: hidden;
+    }
+
+    .stopwatch__hand {
```
lerastarynets, line 19: pls make use of sass nesting for bem selectors everywhere in this file

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

anastasiiavorobiova (approved):
Great work!

PR #3056, opened 2024-01-27
add task solution

valentynavik:
[ x] Keyframes implemented using from/to + transform with rotate property
- [x ] Stopwatch is centered and has correct arrows size
- [ x] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [x ] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)

### DIFF
--- a/src/styles/main.scss
+++ b/src/styles/main.scss
@@ -1,1 +1,0 @@
-// put code here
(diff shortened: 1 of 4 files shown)


etojeDenys (changes requested):
.

src/styles/main.scss
```diff
@@ -1 +1,65 @@
...
+
+@keyframes move {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+}
```
SanyaBratashchuk, line 65: it is better to create a separate file for animations

SanyaBratashchuk (approved):

PR #3057, opened 2024-01-27
stopwatch

emberlast1:
- [DEMO LINK](https://emberlast1.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/mixin.scss
+++ b/src/styles/mixin.scss
@@ -1,0 +1,5 @@
+ @mixin arrows {
+  display: flex;
+  justify-content: center;
+  position: absolute;
+ }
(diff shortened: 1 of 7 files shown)


src/styles/watch.scss
```diff
@@ -0,0 +1,56 @@
+@import './mixin';
+@import './var';
```
TarasHoliuk, line 2: no need to import it here
just import `./watch` after these files in the `main.scss`

src/styles/mixin.scss
```diff
@@ -0,0 +1,5 @@
+ @mixin arrows {
```
TarasHoliuk, line 1:
```suggestion
 @mixin arrow {
```

TarasHoliuk (approved):
GJ :+1: Approved :heavy_check_mark: 

Added 2 tiny comments just FYI

PR #3106, opened 2024-02-12
add task stopwatch

jacube1:
- [DEMO LINK](https://jacube1.github.io/layout_stop-watch/)


yevhenii-pyl (approved):
Good job! 👍 

Definitely not a mistake, but as a quick note here: be more precise if there is a checklist provided to the task. 
You were supposed to  `Copy this Checklist to the Pull Request description after links, and put - [x] before each point after you checked it.`

PR #3123, opened 2024-02-20
add task solution

Ehulinian:
- [DEMO LINK](https://Ehulinian.github.io/layout_stop-watch/)


l4st1m0za (approved):
Good job, but It's better to make so that the hour arrow also moves, but slowlier than the hour arrow. At a pretty good job!

PR #3147, opened 2024-03-01
watch

Nizenko:
[DEMO LINK](https://Nizenko.github.io/layout_stop-watch/)


l4st1m0za (changes requested):
Good job, but you need to make so that the hour-arrow would also move all the time, not only when the minute-hand will make a full rotation

Nizenko:
> Good job, but you need to make so that the hour-arrow would also move all the time, not only when the minute-hand will make a full rotation

Це ж суперечить пункту задачі
> For minutes hand use steps animation with 60 steps. It should take 60min for the minutes hand to make a full circle.

Секундна стрілка лінійно крутиться, а хвилинна покроково (60 steps)

anastasiiavorobiova (approved):
Awesome work!

PR #3158, opened 2024-03-08
stop-watch Zakharchuk Artem

ArtemZakhar:
- [DEMO LINK](https://ArtemZakhar.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,13 +10,21 @@
       http-equiv="X-UA-Compatible"
       content="ie=edge"
     />
+
     <title>Stop watch</title>
+
     <link
       rel="stylesheet"
       href="./styles/index.scss"
     />
   </head>
   <body>
-    <h1>Stop watch</h1>
+    <div class="wrapper">
+      <div class="stopwatch stopwatch--speed-up">
+        <div class="stopwatch__minutes"></div>
+        <div class="stopwatch__seconds"></div>
+        <div class="stopwatch__center"></div>
+      </div>
+    </div>
   </body>
 </html>
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -15,6 +15,7 @@
   height: 80vmin;
   border: 1vmin dotted #000;
   border-radius: 50%;
+  box-sizing: initial;
 
   &__minutes {
     @include hand(20vmin, 3vmin, #0700ff);
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,67 @@
-body {
-  margin: 0;
+@import './reset';
+@import './utils/mixins';
+
+.wrapper {
+  width: 100vw;
+  height: 100vh;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stopwatch {
+  position: relative;
+  width: 80vmin;
+  height: 80vmin;
+  border: 1vmin dotted #000;
+  border-radius: 50%;
+
+  &__minutes {
+    @include hand(20vmin, 3vmin, #0700ff);
+
+    transform: translate(-50%, -10vmin) rotate(-90deg);
+    animation-name: hand-move;
+    animation-duration: 3600s;
+    animation-iteration-count: infinite;
+    animation-timing-function: steps(60);
+  }
+
+  &__seconds {
+    @include hand(38vmin, 1.5vmin, #2c8000);
+
+    transform: translate(-50%, -19vmin) rotate(-90deg);
+    animation-name: hand-move;
+    animation-duration: 60s;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+  }
+
+  &__center {
+    @include hand(5vmin, 5vmin, #f6a603);
+
+    transform: translate(-50%, -1vmin);
+    border-radius: 50%;
+    z-index: 3;
+  }
+
+  &--speed-up {
+    & :nth-child(1) {
+      animation-duration: 600s;
+    }
+
+    & :nth-child(2) {
+      animation-duration: 10s;
+    }
+  }
+}
+
+@keyframes hand-move {
+  from {
+    transform: translate(0, 0) rotate(-90deg);
+    transform-origin: center left;
+  }
+  to {
+    transform: translate(0, 0) rotate(270deg);
+    transform-origin: center left;
+  }
 }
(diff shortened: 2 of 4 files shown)


src/index.html
```diff
@@ -10,13 +10,21 @@
...
     <link
       rel="stylesheet"
       href="./styles/index.scss"
     />
   </head>
   <body>
-    <h1>Stop watch</h1>
+    <div class="wrapper">
+      <div class="stopwatch stopwatch--speed-up">
+        <div class="stopwatch__minutes"></div>
```
mrkovaliv, line 24: Rename `stopwatch__minutes `-> `stopwatch__minute-hand` to improve readability

src/index.html
```diff
@@ -10,13 +10,21 @@
...
+        <div class="stopwatch__seconds"></div>
```
mrkovaliv, line 25: the same)

src/styles/index.scss
```diff
@@ -1,3 +1,68 @@
...
+    @include hand(38vmin, 1.5vmin, #2c8000);
+
+    transform: translate(-50%, -19vmin) rotate(-90deg);
+    animation-name: hand-move;
+    animation-duration: 60s;
+    animation-iteration-count: infinite;
+    animation-timing-function: linear;
+  }
+
+  &__center {
```
mrkovaliv, line 40: You don't need `center element`, you can do the same via `.stopwatch::after`:

```
.stopwatch::after {
  content: '';
  position: absolute;
  top: 50%;
  left: 50%;
  width: 5vmin;
  height: 5vmin;
  background-color: #f6a603;
  border-radius: 50%;
  transform: translate(-50%, -50%);
}
```

src/index.html
```diff
@@ -10,13 +10,21 @@
...
+        <div class="stopwatch__center"></div>
```
mrkovaliv, line 26: you can remove this

src/styles/index.scss
```diff
@@ -1,3 +1,68 @@
...
+    @include hand(5vmin, 5vmin, #f6a603);
+
+    transform: translate(-50%, -1vmin);
+    border-radius: 50%;
+    z-index: 3;
+  }
+
+  &--speed-up {
+    & :nth-child(1) {
```
mrkovaliv, line 49: better use `stopwatch__minute-hand` instead of` :nth-child(1) `because if you will change order of elements or add new element you will need also rewrite css

src/styles/index.scss
```diff
@@ -1,3 +1,68 @@
...
+      animation-duration: 600s;
+    }
+
+    & :nth-child(2) {
```
mrkovaliv, line 53: the same

src/styles/index.scss
```diff
@@ -1,3 +1,68 @@
...
+    animation-name: hand-move;
+    animation-duration: 3600s;
+    animation-iteration-count: infinite;
+    animation-timing-function: steps(60);
+  }
+
+  &__seconds {
+    @include hand(38vmin, 1.5vmin, #2c8000);
+
+    transform: translate(-50%, -19vmin) rotate(-90deg);
```
mrkovaliv, line 33: why did you use -19vmin, may be better -50%?

src/styles/index.scss
```diff
@@ -1,3 +1,68 @@
...
+  width: 80vmin;
+  height: 80vmin;
+  border: 1vmin dotted #000;
+  border-radius: 50%;
+  box-sizing: initial;
+
+  &__minutes {
+    @include hand(20vmin, 3vmin, #0700ff);
+
+    transform: translate(-50%, -10vmin) rotate(-90deg);
```
mrkovaliv, line 23: `-10vmin `may be better -50%?

src/styles/utils/mixins.scss
```diff
@@ -0,0 +1,8 @@
+@mixin hand($width, $height, $bg-color) {
+  position: absolute;
+  top: 49%;
```
mrkovaliv, line 3: why 49%? may be 50%?

src/styles/index.scss
```diff
@@ -1,3 +1,68 @@
...
+  align-items: center;
+}
+
+.stopwatch {
+  position: relative;
+  width: 80vmin;
+  height: 80vmin;
+  border: 1vmin dotted #000;
+  border-radius: 50%;
+  box-sizing: initial;
```
mrkovaliv, line 18: why did you add this style?

mrkovaliv (changes requested):
Nice work, I really like how you used mixins here 👍 🥇 

Added some comments, also try to fix tests before the next review)))

src/index.html
ArtemZakhar: What do you mean by "this"?)
Do you mean to remove stopwatch__center and use ::after pseudo-element?

I will change according to my vision of what you asked)

src/styles/index.scss
ArtemZakhar: ))) 
This is like when you have tests and choose between two options)
First I have made nested classes)
But then I realized that we should not enhance classes specificity according to BEM)

src/styles/index.scss
ArtemZakhar: you can try)
the result may surprise)
first I put this property inside mixin. I don't know how I realized that I should shift a half distance from a width and throw away additional property)
but then I noticed that the center element looks bad and needs to be a little bit lower. So I took the property and passed it inside class. But -50% did not help me) -1200% looked better)))

src/styles/utils/mixins.scss
ArtemZakhar: in case of 50% the hand comes too much close to bottom border and had much free space at the top)

src/styles/index.scss
ArtemZakhar: box-sizing: initial; ?

there was no task to pass tests. But I tried to pass)
and noticed that border-box doesn't suit here)

src/index.html
mrkovaliv: I mean to remove stopwatch__center and use ::after pseudo-element

src/index.html
ArtemZakhar: stopwatch__second-hand looks better, right?))

DanilWeda (approved):
Good. 
I approved. 
But in the future try to use % and px. vh, vw, vmin, and others are used in unique cases, but for the simple layout, you can use a simple structure.
Many thanks!

ArtemZakhar:
@DanilWeda, thank you.
The task was to use vmin.
I wonder if there would be tasks where we have to be using rem and em.

PR #3163, opened 2024-03-11
added solution to stopwatch

vinasxz:
- [DEMO LINK](https://vinasxz.github.io/layout_stop-watch/)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1368,9 +1368,9 @@
       "dev": true
     },
     "@mate-academy/scripts": {
-      "version": "1.7.2",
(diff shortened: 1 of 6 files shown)


jv-aquino (approved):
Good job! A thing to keep in mind for next projects:
Always separate BEM blocks into different css files. A good structure to keep in mind would be something like:
- a file only for imports
- one for global code like body tag and stuff
- one file for every file component

(the default BEM file structure is not that useful until you start using a lot of React, and that's why it's not recommended here, but if you to take a look: https://en.bem.info/methodology/quick-start/#file-structure)

PR #3174, opened 2024-03-13
add solution

hardaira:
https://hardaira.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,71 @@
 body {
   margin: 0;
 }
+
+.clock-face {
+  margin: auto;
+  height: 80vmin;
+  width: 80vmin;
+  border: 1vmin dotted #000;
+  border-radius: 50%;
+  position: relative;
+}
+
+.minutes {
+  height: 20vmin;
+  width: 3vmin;
+  background-color: #0700ff;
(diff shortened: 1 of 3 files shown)


GUSILLUS (changes requested):
Nice job! But let's re-write the code to pass the test

1. We need to use the `.stop-watch` class because npm test needs it
image

2. According to the condition of the problem, the acceleration classes must be in the main class, that is, .stop-watch, and not in the arrows
image

src/styles/index.scss
```diff
@@ -1,3 +1,71 @@
...
+}
+
+.seconds {
+  height: 38vmin;
+  width: 1.5vmin;
+  background-color: #2c8000;
+  position: absolute;
+  top: 3%;
+  left: 50%;
+  margin: -0.75vmin 0 0;
```
TarasHoliuk, line 37: seems strange - and it's not exactly correct position (the reason tests fail)
try to use `flex` - it might help to center your elements

TarasHoliuk: Same for center and minues

TarasHoliuk (changes requested):

OleziO (approved):

PR #3187, opened 2024-03-20
Stopwatch develop

WGoliasz:
- [DEMO LINK](https://WGoliasz.github.io/layout_stop-watch/)

- [ x ] Keyframes implemented using from/to + transform with rotate property
- [ x ] Stopwatch is centered and has correct arrows size
- [ x ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ x ] Code follows all the [Code Style Rules ❗️](https://mate-academy.github.io/layout_task-guideline/html-css-code-style-rules)


src/index.html
```diff
@@ -13,10 +13,20 @@
...
       rel="stylesheet"
-      href="./styles/index.scss"
+      href="./styles/main.scss"
     />
   </head>
   <body>
-    <h1>Stop watch</h1>
+    <div class="stopwatch">
+      <div
+        class="stopwatch__minutes-hand stopwatch__minutes-hand--speed-up"
```
yevhenii-pyl, line 22: Let's shorten it to to make it more readable (here and for other div's)

`  <div class="stopwatch__minutes-hand stopwatch__minutes-hand--speed-up" />`

yevhenii-pyl (changes requested):
Solid: clean and readable! 🚀 
Let's do just a small adjustment in your html file to make it more readable and we are good to go!

WGoliasz:
I'm not sure if I understood correctly, but was it about these changes?
Zrzut ekranu 2024-03-16 133219

yevhenii-pyl (approved):
Exactly!  👍 

Previously you had: 
```
      <div
        class="stopwatch__seconds-hand stopwatch__seconds-hand--speed-up"
      ></div>
```

It's simply not the way html code should look like, there are two mistakes: 

1. You never want to leave a line like this ` **></div>**` 
2. You don't want to pass a class attribute to tag on a separate line if there is just single attribute. 
```
      <div
        **class="stopwatch__seconds-hand stopwatch__seconds-hand--speed-up"**
      >
```
It should be either:

```
      <div
        **==> MORE ATTRIBUTES HERE**
        class="stopwatch__seconds-hand stopwatch__seconds-hand--speed-up"
      >
       **==> SOMETHING HERE**
      </div>
```
In this case you have something inside your div and your element receives multiple params so it makes sense to pass them in separate lines. 

OR, you should shorten it to:

`<div class="stopwatch__minutes-hand stopwatch__minutes-hand--speed-up"</div>
`

And since div is a self-closing tag, you can shorten it even more this way: 
`<div class="stopwatch__minutes-hand stopwatch__minutes-hand--speed-up" />`

PR #3197, opened 2024-03-26
stop watch solution

itsmeoleg:
- [DEMO LINK] (https://itsmeoleg.github.io/layout_stop-watch/)

- [ x ] Keyframes implemented using from/to + transform with rotate property
- [ x ] Stopwatch is centered and has correct arrows size
- [ x ] All `Typical Mistakes` from `BEM` lesson theory are checked.
- [ x ] Code follows all the


loralevitska (approved):
Well done)

PR #3239, opened 2024-04-23
create stop-watch

Misha20850:
- [DEMO LINK](https://Misha20850.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/block/watch.scss
+++ b/src/styles/block/watch.scss
@@ -1,0 +1,78 @@
+body,
+html {
+  height: 100%;
+  margin: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stopwatch {
+  width: 80vmin;
+  height: 80vmin;
+  border: 1vmin dotted #000;
+  border-radius: 50%;
+  position: relative;
+}
+
+.hand {
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform-origin: 50% 100%;
+
+  &--minutes {
+    width: 3vmin;
+    height: 20vmin;
+    background-color: #0700ff;
+    animation: rotate-minutes 3600s steps(60) infinite;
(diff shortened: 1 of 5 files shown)


src/styles/block/watch.scss
```diff
@@ -0,0 +1,78 @@
...
+}
+
+@keyframes rotate-seconds {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
```
vadiimvooo, line 78: You don't need to create two separate keyframes as animation is the same

src/styles/block/watch.scss
```diff
@@ -0,0 +1,78 @@
...
+  left: 50%;
+  transform-origin: 50% 100%;
+
+  &--minutes {
+    width: 3vmin;
+    height: 20vmin;
+    background-color: #0700ff;
+    animation: rotate-minutes 3600s steps(60) infinite;
+    top: calc(50% - 20vmin);
+    margin-left: -1.5vmin;
```
vadiimvooo, line 30: Why did you use such strange margin. You can use translate property in your keyframe transform.

src/styles/block/watch.scss
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  &--seconds {
+    width: 1.5vmin;
+    height: 38vmin;
+    background-color: #2c8000;
+    animation: rotate-seconds 60s linear infinite;
+    top: calc(50% - 38vmin);
+    margin-left: -0.75vmin;
```
vadiimvooo, line 39: The same is for seconds hand

vadiimvooo (changes requested):
Good job, but there is a few changes needs to be fixed.

etojeDenys (approved):
good job

PR #3252, opened 2024-05-01
add solution

Razsinxron:
- [DEMO LINK](https://Razsinxron.github.io/layout_stop-watch/)

- [ ] Keyframes implemented using from/to + transform with rotate property

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,77 @@
+@mixin styles($color, $height, $width) {
+  background-color: $color;
+  height: $height;
+  width: $width;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform-origin: 50% 100%;
+}
+
+@mixin stopwatch-styles {
+  position: relative;
+  height: 80vmin;
+  width: 80vmin;
+  border: dotted 1vmin #000;
+  border-radius: 50%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
 body {
   margin: 0;
+  height: 100vh;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.stopwatch {
+  @include stopwatch-styles;
+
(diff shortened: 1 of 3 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,77 @@
...
+
+  &__second {
+    @include styles(#2c8000, 38vmin, 1.5vmin);
+
+    transform: translate(-50%, -100%);
+    animation: hand-move 60s linear infinite;
+  }
+
+  &--speed-up {
+    @include stopwatch-styles;
```
TarasHoliuk, line 56: Redundant - these styles exists in `.stopwatch` so we don't need to add them again with `stopwatch--speed-up`, only change animation property for clockwatch hands

src/styles/index.scss
```diff
@@ -1,3 +1,77 @@
...
+  background-color: $color;
+  height: $height;
+  width: $width;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform-origin: 50% 100%;
+}
+
+@mixin stopwatch-styles {
```
TarasHoliuk, line 11: should be used once, so makes not a lot of sence

src/styles/index.scss
```diff
@@ -1,3 +1,77 @@
+@mixin styles($color, $height, $width) {
```
TarasHoliuk, line 1:
```suggestion
@mixin hand-styles($color, $height, $width) {
```
good idea, but let's rename it

TarasHoliuk (changes requested):
GJ 👍 
A few small comments only

src/styles/index.scss
Razsinxron: If the dial should remain, it should not be removed, it will not be displayed without this style

volodymyr-soltys97 (approved):
Good job 👍

PR #3292, opened 2024-05-26
taskUpload

papayapan:
https://papayapan.github.io/layout_stop-watch/

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,86 @@
-body {
+@import '../utils/variables.scss';
+
+body, html {
+  height: 100%;
   margin: 0;
+  padding: 0;
+  @include flex($justify: center, $align:center)
 }
+
+
+.stopwatch{
+  @include flex($justify: center, $align:center);
+  height: 100%;
+  width: 100%;
+  &__container{
+    border-style:dotted;
+    border-color: #000;
+    border-width: 1vmin;
+    border-radius: 50%;
+    width: 80vmin;
+    height: 80vmin;
+    position: relative;
+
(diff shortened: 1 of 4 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,86 @@
...
+}
+
+@keyframes rotate-minutes {
+  from {
+      transform: rotate(0deg);
+  }
+  to {
+      transform: rotate(360deg);
+  }
+}
```
vadiimvooo, line 75: You don't need to do 2 animations, just use one of them

src/styles/index.scss
```diff
@@ -1,3 +1,86 @@
...
+
+.stopwatch--speed-up .stopwatch__container_secondsHands {
+  animation-duration: 10s;
+}
+
+.stopwatch-speed-up .stopwatch__container_minutesHands {
+  animation-duration: 600s;
+}
+
+
```
vadiimvooo, line 85: Please add modifiers to increase speed of minutes and seconds hand.

src/styles/index.scss
```diff
@@ -1,3 +1,86 @@
...
+    width: 80vmin;
+    height: 80vmin;
+    position: relative;
+
+    @include flex($justify: center, $align:center);
+    &_circle{
+      position: relative;
+      width: 5vmin;
+      height: 5vmin;
+      background-color: #f6a603;
```
vadiimvooo, line 29: use variable to store color value

vadiimvooo: The same in other places

vadiimvooo (changes requested):
Good job, there are a few changes.

PR #3296, opened 2024-05-28
add task solution

merzotnik:
- [DEMO LINK](https://merzontik.github.io/layout_stop-watch/)
- [x] Code follows all the [Code Style Rules ❗️]

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,108 @@
 body {
   margin: 0;
 }
+
+.container {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
+}
+
+.watch {
+  position: relative;
+  height: 80vmin;
+  width: 80vmin;
+  content: '';
+  border: 1vmin dotted #000;
+  border-radius: 100%;
+}
+
+.arrow-min {
+  position: absolute;
+  bottom: 50%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  content: '';
+  height: 20vmin;
+  width: 3vmin;
+  background-color: #0700ff;
+  transform-origin: bottom;
+  animation: min 3600s linear infinite;
+}
+
+.arrow-sec {
+  position: absolute;
+  bottom: 50%;
+  z-index: 1;
+  display: flex;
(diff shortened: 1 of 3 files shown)


volodymyr-soltys97 (changes requested):
Hi, your demo link isn't working, you need to make a deploy and check it after on the demo link 
image

volodymyr-soltys97 (changes requested):
Still not fixed from the previous review, if you have some problems with deploy - feel free  in the fe_chat 
image

src/styles/index.scss
```diff
@@ -1,3 +1,108 @@
...
+
+@keyframes min {
+  from {
+    transform: rotate(0deg);
+  }
+
+  to {
+    transform: rotate(360deg);
+  }
+}
```
vadiimvooo, line 108: This animations are the same. Please use one of them

src/styles/index.scss
```diff
@@ -1,3 +1,108 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  content: '';
+  height: 20vmin;
+  width: 3vmin;
+  background-color: #0700ff;
+  transform-origin: bottom;
+  animation: min 3600s linear infinite;
+}
```
vadiimvooo, line 33: You have doublicated properties which you reuse in different selectors. To remove doublication you can use [scss placeholders](https://sass-lang.com/documentation/style-rules/placeholder-selectors/)

vadiimvooo: Please fix it for all of the selectors below

vadiimvooo (changes requested):
Good job. There are a few comments to fix.

merzotnik:
I don't understand why my test doesn't pass after running the npm test command, but if I intentionally make a mistake, run the test, then fix the mistake, the test passes without any problems. The only value I'm changing is the animation-duration. Its initial value is 600s, which is correct. I intentionally set it to 599s, and the test doesn't pass in both cases. Then I change it back to 600s, and the test passes. What could be the problem?

volodymyr-soltys97 (approved):
Good job 👍

PR #3324, opened 2024-06-28
Develop Stopwatch

Kimmladshii:
- [DEMO LINK](https://kimmladshii.github.io/layout_stop-watch/) 

### DIFF
--- a/src/styles/_stopwatch.scss
+++ b/src/styles/_stopwatch.scss
@@ -1,0 +1,71 @@
+main {
+  height: 100vh;
+
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.stopwatch {
+  height: 80vmin;
+  width: 80vmin;
+  border: 1vmin dotted #000;
+  border-radius: 50vmin;
+  position: relative;
+
+  &__minutes {
+    height: $minute-arrow-height;
+    width: $minute-arrow-width;
+    background-color: $minute-arrow-color;
+
+    @include position-absolute-center;
+
+    z-index: 1;
+
+    transform-origin: bottom;
+    animation: watch-start steps(60) infinite;
+  }
+
+  &__seconds {
+    height: $seconds-arrow-height;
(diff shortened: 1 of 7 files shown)


BogdanMaliuta (changes requested):
Hello, code looks good and works good.But if I removing modificators t`s doesnt work correctly

src/styles/_stopwatch.scss
```diff
@@ -0,0 +1,71 @@
+main {
+  height: 100vh;
+
```
Zibi95, line 3: Are you leaving those empty lines for purpose?

src/styles/_stopwatch.scss
```diff
@@ -0,0 +1,71 @@
...
+  &__pivot {
+    height: $pivot-height;
+    width: $pivot-width;
+    background-color: $pivot-color;
+    border-radius: 50vmin;
+
+    @include position-absolute-center;
+
+    z-index: 3;
+    transform: translate(-50%, -50%);
```
Zibi95, line 51: I think with display flex it would be better.

src/styles/_stopwatch.scss
```diff
@@ -0,0 +1,71 @@
...
+    height: $seconds-arrow-height;
+    width: $seconds-arrow-width;
+    background-color: $seconds-arrow-color;
+
+    @include position-absolute-center;
+
+    z-index: 2;
+
+    transform-origin: bottom;
+    animation: watch-start 10s linear infinite;
```
Zibi95, line 39: Here should be 60s. You can also make variables with seconds and then overwrite the values inside selector with modifier.

Zibi95 (changes requested):
It still doesn't work  as it should after I remove modifier.

Kimmladshii:
> It still doesn't work as it should after I remove modifier.

I removed the modifier and nothing breaks.

natalia-klonowska (approved):

PR #3343, opened 2024-07-18
Stopwatch animation.

kamilbykow:
[DEMO LINK](https://kamilbykow.github.io/layout_stop-watch/)


src/index.html
```diff
@@ -13,10 +13,16 @@
     <title>Stop watch</title>
     <link
       rel="stylesheet"
-      href="./styles/index.scss"
+      href="./styles/index.css"
```
natalia-klonowska, line 16: you should use scss file

natalia-klonowska (changes requested):
Great work but you don't need to commit css and css.map files

PR #3348, opened 2024-07-22
Develop

Miranna472:

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -54,7 +54,7 @@ html {
 }
 
 .stopwatch__hand--second {
-  width: 2vmin;
+  width: 1.5vmin;
   height: 38vmin;
   background: #2c8000;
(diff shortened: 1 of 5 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,97 @@
...
+  }
+}
+
+@keyframes rotate-second {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
```
lerastarynets, line 96: pls combine these 2 animations to 1, because they do the same thing and reuse where needed

lerastarynets (approved):
I see you did a hard work) Let's improve it a little bit

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

PR #3379, opened 2024-08-15
add stopwatch solution

modeltoIT:
- [DEMO LINK](https://modeltoIT.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/utils/animations.scss
+++ b/src/styles/utils/animations.scss
@@ -1,0 +1,19 @@
+@keyframes rotate-minute {
+  from {
+    transform: translate(-50%, 0) rotate(0);
+  }
+
+  to {
+    transform: translate(-50%, 0) rotate(360deg);
+  }
+}
+
+@keyframes rotate-second {
+  from {
+    transform: translate(-50%, -47.5%) rotate(0);
+  }
+
+  to {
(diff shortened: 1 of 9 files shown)


src/styles/utils/animations.scss
```diff
@@ -0,0 +1,19 @@
...
+
+@keyframes rotate-second {
+  from {
+    transform: translate(-50%, -47.5%) rotate(0);
+  }
+
+  to {
+    transform: translate(-50%, -47.5%) rotate(360deg);
+  }
+}
```
volodymyr-soltys97, line 19: The rotation script is the same, so you only need to use one
```suggestion
@keyframes rotate {
  from {
    transform: rotate(0);
  }

  to {
    transform: rotate(360deg);
  }
}
```

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's make your code better

src/styles/utils/animations.scss
modeltoIT: > The rotation script is the same, so you only need to use one

Since we remove 'translate' property with the new keyframe, all the positioning becomes wrong, that's why I continue using "translate" property along with "rotate".

lerastarynets (approved):
lgtm 🚀

PR #3386, opened 2024-08-16
Task Solution "Stop Watch"

MateuszCieplak:
[DEMO LINK] (https://MateuszCieplak.github.io/layout_stop-watch/)


Zibi95 (changes requested):
Hey. Everything is ok but still missing one thing from req

> For minutes hand use steps animation with 60 steps. It should take 60min for the minutes hand to make a full circle.
In addition to the basic functionality create a BEM modifier called speed-up for your stopwatch block:

> it will take only 10s for the seconds hand to make a full circle (change animation duration)
> it will take only 10min for the minutes hand to make a full circle (change animation duration)

Zibi95 (approved):
Good job! 🥇

PR #3392, opened 2024-08-19
add solution

VladPopenko:
- [DEMO LINK](https://VladPopenko.github.io/layout_stop-watch/)

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,6 +17,10 @@
     />
   </head>
   <body>
-    <h1>Stop watch</h1>
+    <div class="stopwatch">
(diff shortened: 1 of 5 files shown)


src/index.html
```diff
@@ -17,6 +17,10 @@
     />
   </head>
   <body>
-    <h1>Stop watch</h1>
+    <div class="stopwatch">
```
volodymyr-soltys97, line 20: Add the modifier class here 
image
```suggestion
    <div class="stopwatch stopwatch--speed-up">
```

volodymyr-soltys97 (changes requested):
Good job!
Need to fix UI after adding the modifier classes 
image

PR #3418, opened 2024-09-13
changed firstly

Naathie:
- [DEMO LINK](https://Naathie.github.io/layout_stop-watch/)


src/styles/index.scss
```diff
@@ -1,3 +1,89 @@
...
+}
+
+@keyframes rotate-seconds {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
 }
```
volodymyr-soltys97, line 89: Use only one keyframe, because the scenario is the same

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code
This part of the task is incomplete, you need to fix time rotate 
image

volodymyr-soltys97 (approved):
Good job! 👍

PR #3443, opened 2024-10-04
add task solution

izotovav:
- [DEMO LINK](https://izotovav.github.io/layout_stop-watch/)

### DIFF
--- a/src/styles/index.scss
+++ b/src/styles/index.scss
@@ -1,3 +1,79 @@
+$hand-transform-origin: bottom center;
+$hand-bottom: 50%;
+
 body {
   margin: 0;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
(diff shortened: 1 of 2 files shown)


src/styles/index.scss
```diff
@@ -1,3 +1,79 @@
...
+@keyframes rotate-minute {
+  from {
+    transform: rotate(0deg);
+  }
+  to {
+    transform: rotate(360deg);
+  }
+}
+
+@keyframes rotate-second {
```
VitaliyBondarenko1982, line 72: why do you need same keyframes?
looks like you can create just one keyframe and use in two cases.

VitaliyBondarenko1982 (changes requested):
Good job.
Please check comment

anastasiiavorobiova (approved):
Awesome!

PR #4102, opened 2026-03-19
add stopwatch layout

linska:


linska:
The current setup uses left bottom + translateX(-50%), so the element is visually re-centered before rotation. Because of that, the pivot point is effectively correct even if the origin value looks unconventional.

The resulting animation matches the expected behavior.

PR #4107, opened 2026-03-20
implement stopwatch layout

sonik-boom71:
DEMO LINK
https://sonik-boom71.github.io/layout_stop-watch/

- [x] Stopwatch is centered on the page
- [x] Stopwatch uses only CSS animations
- [x] Second hand rotates in 60s with linear timing
- [x] Minute hand rotates in 60min with steps(60)
- [x] speed-up modifier works correctly
- [x] backstopConfig.js uses my own GitHub Pages URL


brespect (changes requested):
Good job, but lets fix all tests beore requesting the review

PR #4204, opened 2026-08-03
Implement stopwatch with CSS animations

luanstipp2018-arch:
Rotate minute and second hands via keyframes + transform, with a speed-up BEM modifier that shortens both animation durations.


luanstipp2018-arch (commented):
...

luanstipp2018-arch (commented):
...

